PR #2, opened 2020-01-30
feat: add landing page (LFA-1259)

geek-soft:


src/controllers/HomeController.ts
```diff
@@ -2,5 +2,5 @@ import { Request, Response, NextFunction } from 'express';
 
 export class HomeController {
 
-    public sayHello = (req: Request, res: Response, next: NextFunction) => res.render('index');
+    public sayHello = (req: Request, res: Response, next: NextFunction) => res.render('landing');
```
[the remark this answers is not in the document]
geek-soft: Of course `sayHello` is not great name. If we say we will have one controller / route per path / page then maybe we can use universal names such as `getHandler` and `postHandler` or `onGetRequest` and `onPostRequest`. Of course nothing stops us from having other private methods inside same controller / route.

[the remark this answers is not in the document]
geek-soft: @n-r-b , @franciscog-kainos what do you think about suggested names for methods handling http requests?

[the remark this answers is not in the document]
franciscog-kainos: The way I see it we should have all the functions that handle, for example the 'Landing Page', inside a landing controller, etc...

akaJSaunders:
I have pushed two commits to the landing page branch. 
1. Created the common layout view to which all other pages should extend from although individual pages have to import gov uk components that are not covered by the basic layout template.
2. Fixed the missing blue component on the subsection. As it wasn't a gov-uk component, had to create a _related-items.scss and imported it from the application.scss.

src/controllers/HomeController.ts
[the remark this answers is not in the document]
n-r-b: what about 'getLandingPage' if this function is returning the landing page or 'renderLandingPage' if the return type is a void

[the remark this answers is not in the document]
geek-soft: I personally prefer generic names cause these controllers will always be easy with smaller private functions doing heavy lifting.

src/views/landing.njk
```diff
@@ -0,0 +1,83 @@
+{% extends "layout.njk" %}
+{% extends "govuk/template.njk" %}
```
geek-soft: I bet we do not need that since `layout.njk` extends `govuk/template.njk` as well.

src/public/sass/patterns/_related-items.scss
```diff
@@ -0,0 +1,12 @@
+// This is a GOV.UK Publishing specific component that
+// can be seen at http://govuk-static.herokuapp.com/component-guide/related_items
```
geek-soft: The link did not work for me. Shouldn't that be https://components.publishing.service.gov.uk/component-guide/related_navigation?

src/views/landing.njk
```diff
@@ -0,0 +1,83 @@
...
+
+{% from "govuk/components/back-link/macro.njk"   import govukBackLink %}
+{% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
+{% from "govuk/components/button/macro.njk"      import govukButton %}
+
+{% block pageTitle %}
+  {% if serviceName %}
```
geek-soft: I suggest we remove that if statement and set service name globally

akaJSaunders: The service name is set in the layout currently, I agree that if statement is not necessary

src/public/sass/patterns/_related-items.scss
akaJSaunders: I don't think that is the correct link. the comments were taken from the prototype and since are out of date it is probably worth deleting.

src/views/landing.njk
akaJSaunders: Well spotted

src/views/layout.njk
```diff
@@ -42,22 +28,29 @@
   {% block download %}{% endblock %}
 {% endblock %}
 
+
+
+{% block pageTitle %}
```
geek-soft: I do not think it is needed. Layout is for extension so usually extending pages will declare it.

src/views/layout.njk
```diff
@@ -86,6 +79,7 @@
   }) }}
 {% endblock %}
 
+
```
geek-soft: I think we should remove that line, change itself bring no value and just makes diff bigger.

src/views/layout.njk
```diff
@@ -1,27 +1,13 @@
...
-{% from "govuk/components/button/macro.njk" import govukButton %}
-{% from "govuk/components/date-input/macro.njk" import govukDateInput %}
-{% from "govuk/components/details/macro.njk" import govukDetails %}
-{% from "govuk/components/error-message/macro.njk" import govukErrorMessage %}
-{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
-{% from "govuk/components/footer/macro.njk" import govukFooter %}
-{% from "govuk/components/header/macro.njk" import govukHeader %}
-{% from "govuk/components/input/macro.njk" import govukInput %}
-{% from "govuk/components/panel/macro.njk" import govukPanel %}
+{% extends "govuk/template.html" %}
```
geek-soft: It should be `{% extends "govuk/template.njk" %}` here.

src/views/landing.njk
```diff
@@ -0,0 +1,83 @@
...
+  </div>
+{% endblock %}
+
+{% block content%}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl">Appeal a late filing penalty</h1>
+      <p class="govuk-body-l">Use this service to appeal a penalty issued to a company for not filing its annual accounts on time.</p>
+      <p>It may take up to 30 minutes to complete this application.</p>
+      {{ govukButton({
```
geek-soft: Not big deal but what if we had the following formatting?

image

akaJSaunders: agreed

src/views/layout.njk
akaJSaunders: silly mistake of mine

src/views/landing.njk
```diff
@@ -9,26 +8,21 @@
...
-      <li class="govuk-breadcrumbs__list-item">
-        <a class="govuk-breadcrumbs__link" href="https://www.gov.uk/browse/business">Business and self-employed</a>
-      </li>
-      <li class="govuk-breadcrumbs__list-item">
-        <a class="govuk-breadcrumbs__link" href="https://www.gov.uk/browse/business/limited-company">Running a limited company</a>
-      </li>
-    </ol>
-  </div>
+  {{
+    govukBreadcrumbs({
```
geek-soft: I suggest we use macros as much as possible creating our own if needed. They nicely hide CSS complexity and make us focus on content, not mentioning reusability.

src/views/landing.njk
```diff
@@ -9,26 +8,21 @@
 {% endblock %}
 
 {% block beforeContent %}
-  {% block backLink %}{% endblock %}
```
geek-soft: I Decided to remove it as it is not used.

src/views/landing.njk
```diff
@@ -42,7 +36,7 @@
       <p>
         <a href="https://www.gov.uk/government/publications/late-filing-penalties/late-filing-penalties#appealing" target="_blank">Read the guidance on appealing a penalty.</a>
       </p>
-      <h3 class="govuk-heading-s">You'll need:</h3>
+      <h3 class="govuk-heading-s">You’ll need:</h3>
```
geek-soft: Dedicated apostrophe char renders bit nicer UI and it does not break template syntax highlight. I suggest we use it in cases like that instead of using single quote.

How to achieve it? It is `option+shift+]` on Mac.

PR #3, opened 2020-01-30
WIP: feat: started working on redis and DI

franciscog-kainos:
To make the frontend code seem more familiar and easier to use, I've added a DI library (https://github.com/inversify/inversify-express-utils) to the project.

I've also started playing around with the redis client. 

To properly run this:
- Create a redis docker instance with `docker run --name redis-instance -p 6379:6379 -d redis`
- `npm clean-install` for good measure
- `npm start`


tsconfig.json
```diff
@@ -10,6 +10,8 @@
       "emitDecoratorMetadata": true,
       "moduleResolution": "node",
       "importHelpers": true,
+      "lib": ["es6", "dom"],
```
geek-soft: What do we need `dom` for?

package.json
```diff
@@ -27,16 +28,20 @@
...
+    "inversify": "^5.0.1",
+    "inversify-express-utils": "^6.3.2",
+    "gulp": "^4.0.2",
     "gulp-sass": "^4.0.2",
     "gulp-typescript": "^6.0.0-alpha.1",
     "gulp-util": "^3.0.8",
     "gulp-watch-sass": "^1.4.1",
     "node-sass": "^4.13.1",
     "sass-lint": "^1.13.1",
+    "reflect-metadata": "^0.1.13",
```
geek-soft: That is more runtime dependency, lack of it will stop an app from starting. I suggest you move it to a block above.

package.json
```diff
@@ -27,16 +28,20 @@
     "@types/express": "^4.0.37",
     "@types/express-session": "^1.15.3",
     "@types/nunjucks": "^3.1.3",
+    "@types/redis": "^2.8.14",
     "del": "^5.1.0",
-    "gulp": "^4.0.0",
+    "inversify": "^5.0.1",
```
geek-soft: I think it should not be in dev dependencies cause DI is used at runtime. You can test it by removing node modules, running `npm install --production` and finally starting the app.

src/utils/AppUtils.ts
```diff
@@ -0,0 +1,10 @@
+export const LogPipe = <A>(log: (a: A) => void) => (a: A) => {
```
geek-soft: These are not being used so please remove unless used.

tsconfig.json
franciscog-kainos: I was testing reloading the page

franciscog-kainos:
I've uploaded the penalty reference screen. Still with no behaviour. I also had to change the css and assets paths in layout.njk (previously index.njk) to become absolute paths.
i.e. if you go to `localhost:3000/index` it uses `/css/etc.css` instead of `/index/css/etc.css`

inversify.config.ts
```diff
@@ -0,0 +1 @@
+import 'reflect-metadata';
```
n-r-b: is this needed?

src/Server.ts
```diff
@@ -1,46 +1,59 @@
...
-    this.app.listen(this.app.get('port'), () => {
-      console.log(('  App is running at http://localhost:%d in %s mode'), this.app.get('port'), this.app.get('env'));
+    this.server.build().listen(this.port, () => {
+      console.log(('  App is running at http://localhost:%d in %s mode'), this.port, process.env.NODE_ENV);
       console.log('  Press CTRL-C to stop\n');
     });
   }
 
-  private setupStaticFolders(app: express.Express): void {
+  private setupBindings(): Container {
```
n-r-b: I feel like this method name is a little misleading, it suggests we are only setting up bindings, but we are actually creating a new container, configuring it's bindings and then returning the container. Instead we could create the Container outside of the method and pass it in as a parameter?

src/Server.ts
```diff
@@ -1,46 +1,59 @@
...
-    this.setupParsers(this.app);
-    this.setViewEngine(this.app);
-    this.setupAppConfig(this.app);
   }
 
   public start(): void {
-    this.app.listen(this.app.get('port'), () => {
-      console.log(('  App is running at http://localhost:%d in %s mode'), this.app.get('port'), this.app.get('env'));
+    this.server.build().listen(this.port, () => {
+      console.log(('  App is running at http://localhost:%d in %s mode'), this.port, process.env.NODE_ENV);
```
n-r-b: where is 'NODE_ENV' set?

src/Server.ts
```diff
@@ -1,46 +1,59 @@
...
 import * as cookieParser from 'cookie-parser';
 import * as express from 'express';
-import * as session from 'express-session';
 import * as nunjucks from 'nunjucks';
+import 'reflect-metadata';
 import * as path from 'path';
-import { router } from './routes/Routes';
+import { Container } from 'inversify';
+import { InversifyExpressServer } from 'inversify-express-utils';
+import './controllers/index';
```
n-r-b: How do we plan to use this?

inversify.config.ts
geek-soft: Good question. I did not figure out how DI framework works.

src/controllers/HomeController.ts
```diff
@@ -1,6 +1,15 @@
...
-export class HomeController {
+@controller('/')
+export class HomeController extends BaseHttpController {
 
-    public sayHello = (req: Request, res: Response, next: NextFunction) => res.render('index');
+    constructor() {
+        super();
+    }
+
+    @httpGet('index')
```
geek-soft: IMO a bit to much is going on in that PR. It has DI, Redis and new page. I'd be cautious before adding more and more, we already have 15 files touched.

src/services/RedisService.ts
```diff
@@ -0,0 +1,14 @@
+import { createClient, RedisClient } from 'redis';
+import { injectable } from 'inversify';
+
+@injectable()
+export class RedisService {
+    public createNewSession(): RedisClient {
```
geek-soft: I suggest we remove word session from method name. This service has nothing to do with session despite session handling being one possible use case. As far as this service is concerned its responsibility is creation of redis client.

tsconfig.json
geek-soft: Does it mean we do not need it anymore? If so remove.

tsconfig.json
```diff
@@ -10,12 +10,15 @@
...
+      "types": ["reflect-metadata"],
       "plugins": [
         {
           "name": "typescript-tslint-plugin"
         }
       ]
     },
+    "experimentalDecorators": true,
```
geek-soft: It is already defined at line 8. Please remove.

inversify.config.ts
franciscog-kainos: This import is needed for the decorators to work. It allows metadata to be added to the objects

src/Server.ts
franciscog-kainos: I'll change this to `createContainerWithBindings(): Container` because creating a container outside and passing it in as a parameter only to then mutate it's state is a side effect we can avoid ;)

src/Server.ts
franciscog-kainos: Not set atm but in the future it will either be set as an environment variable or using the dotenv lib

src/Server.ts
franciscog-kainos: This import just imports all the controllers inside the index so that we don't have to mention them all in this file. These also need to be imported once in this file so the Dependency Resolver can resolve the dependencies.

src/services/RedisService.ts
franciscog-kainos: Changed it to `createClient`

inversify.config.ts
franciscog-kainos: image

src/controllers/HomeController.ts
n-r-b: this is more of a 'set up' ticket for redis and a playground for DI - that was my understanding. We plan to do the work for the ticket on a separate branch

src/Server.ts
n-r-b: do we have to have an index for this? can we not just point to the '/controllers' directory

src/controllers/HomeController.ts
geek-soft: Good question, I am bit sad how this PR grows and by the fact extra bits like DI are included although I understand Francisco is working his way to establish some patterns.

Sadly it will grow even further when tests are added. I would not mind extracting some bits to smaller PR's if that simple and does not add unnecessary time.

src/controllers/index.ts
```diff
@@ -0,0 +1,3 @@
+import './HomeController';
+import './PenaltyReferenceController';
+import './LandingController';
```
akaJSaunders: What is the purpose of this file?

src/views/index.njk
```diff
@@ -0,0 +1,2 @@
+{% extends "layout.njk" %}
+<h1>Hello there</h1>
```
akaJSaunders: similar to another comment of mine, but what is the purpose of index.njk?

src/views/penalty-reference.njk
```diff
@@ -0,0 +1,121 @@
...
+
+{% block beforeContent %}
+  {{ govukPhaseBanner({
+    tag: {
+      text: "alpha"
+    },
+    html: 'This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.'
+    }) 
+  }}
+  {{
```
akaJSaunders: I realize this was actually my work, but looking back at it maybe we should just refactor it so that it looks like:
{{
   govukPhaseBanner {{
    ..... 
   })
}}
Damian had made the suggestion to me in lines 38 to 49 of this same file and it looks much better

PR #8, opened 2020-01-31
Skeleton/dockerized

franciscog-kainos:
Created a Dockerfile and docker-compose.yml which contains redis and the app. This is just proof of concept it only took me 20 minutes anyway I don't expect this to get merged


geek-soft:
If I was in client shoes I would not get enthusiastic hearing about some work that might not be merged, even if this is just 20 minutes.

I'd love to see benefits of Docker but it is big unknown right now unless we establish proper use case of it. It is definitely not needed locally unless the aim is to remove need to install node on dev machines.

Closing it for now without removing branch - might be reopened later.

PR #9, opened 2020-01-31
refactor: Refactored controllers to use DI library

franciscog-kainos:
This PR just adds DI to the project.


package.json
```diff
@@ -34,8 +41,8 @@
     "gulp-watch-sass": "^1.4.1",
     "node-sass": "^4.13.1",
     "sass-lint": "^1.13.1",
-    "tslint": "^6.0.0",
-    "typescript": "^3.7.5",
+    "tslint": "^5.7.0",
```
geek-soft: Please revert that change please. I suggest we use latest versions.

package.json
```diff
@@ -34,8 +41,8 @@
...
+    "typescript": "^3.5.3",
```
geek-soft: Same here

src/controllers/LandingController.ts
```diff
@@ -1,6 +1,11 @@
 import { Request, Response, NextFunction } from 'express';
+import { controller, httpGet } from 'inversify-express-utils';
 
+@controller('/')
 export class LandingController {
 
-    public renderView = (req: Request, res: Response, next: NextFunction) => res.render('landing');
+    @httpGet('')
```
geek-soft: Out of curiosity is that empty string required or can that arg be omitted?

franciscog-kainos: It's required :(

tsconfig.json
```diff
@@ -10,6 +10,8 @@
       "emitDecoratorMetadata": true,
       "moduleResolution": "node",
       "importHelpers": true,
+      "lib": ["es6"],
+      "types": ["reflect-metadata"],
```
geek-soft: After reading https://www.typescriptlang.org/v2/en/tsconfig#types I suggest we remove it. It narrows down what types are loaded if I understand correctly.

src/views/layout.njk
```diff
@@ -82,6 +82,6 @@
 
 {% block bodyEnd %}
   {# Run JavaScript at end of the <body>, to avoid blocking the initial render. #}
-  <script src="govuk/all.js"></script>
+  <script src="/govuk/all.js"></script>
```
geek-soft: Revert to keep it focused on DI. Can be done in PR for user story.

src/Server.ts
```diff
@@ -1,46 +1,58 @@
...
-    this.app.listen(this.app.get('port'), () => {
-      console.log(('  App is running at http://localhost:%d in %s mode'), this.app.get('port'), this.app.get('env'));
+    this.server.build().listen(this.port, () => {
+      console.log(('  App is running at http://localhost:%d in %s mode'), this.port, process.env.NODE_ENV);
       console.log('  Press CTRL-C to stop\n');
     });
   }
 
-  private setupStaticFolders(app: express.Express): void {
+  private createContainerWithBindings(): Container {
```
geek-soft: Good for now but in the future it should be moved away from server.ts to reduce reasons for change.

src/Types.ts
```diff
@@ -0,0 +1,2 @@
+export const TYPES = {
```
n-r-b: do we need to add this right now?

PR #10, opened 2020-01-31
Redis

franciscog-kainos:
PR containing the scafolding for redis


src/services/RedisService.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@injectable()
+export class RedisService {
+
+    public readonly client: RedisClient;
+    constructor() {
+        this.client = this.createClient();
+    }
+
+    private createClient(): RedisClient {
```
n-r-b: I don't feel the redis client configuration belongs here in the service layer, we could create the client in the main server.js or somewhere similar? Also, could we inject the host/port properties from a config file too?

franciscog-kainos: Agreed, we could make the RedisClient injectable as there is ever only going to be one instance of it?

n-r-b: Yes we could! Just seen an example here -> https://github.com/inversify/inversify-express-example/tree/master/BindingDecorators/utils/mongodb

n-r-b: and then we could inject the client like here -> https://github.com/inversify/inversify-express-example/blob/master/BindingDecorators/service/user.ts

README.md
```diff
@@ -22,6 +22,7 @@ To run this app, cd to the root directory, and type the following commands:
 - `npm start`
 
 - Create a redis docker instance with `docker run --name redis-instance -p 6379:6379 -d redis`
+- Copy and rename the `.env.example` file to `.env.development` and change the variables accordingly. For current setup: `REDIS_HOST=127.0.0.1` and `REDIS_PORT=6379`
```
n-r-b: I like the sound of different config files per environment, but how about a generic config file for this task? Not really sure what the plans are with environment variables, something to discuss with @geek-soft

src/utils/RedisUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+export const setAsync = (client: RedisClient) => promisify(client.set).bind(client);
+export const getObjectAsync = (client: RedisClient) => promisify(client.hgetall).bind(client);
+export const setObjectAsync = (client: RedisClient) => async <T>(key: string, values: IMap<T>) =>
+    new Promise((resolve, reject) => {
+        client.hmset(key, values as any, (err, reply) => {
+            if (err) return reject(err);
+            return resolve(reply);
+        });
+    });
+export const createRedisClient = () => createClient({
```
n-r-b: I can see that this code is in two places - here and the server.ts. Does it need to be?
Also if the connection was to fail or timeout, what happens then?

src/middleware/SessionMiddleware.ts
```diff
@@ -0,0 +1,29 @@
+import { BaseMiddleware } from 'inversify-express-utils';
+import { Request, Response, NextFunction } from 'express';
+import { injectable, inject } from 'inversify';
+import { RedisService } from '../services/RedisService';
+import { TYPES } from '../Types';
+
+@injectable()
+export class SessionMiddleware extends BaseMiddleware {
```
n-r-b: Session management isn't part of this task. Keep it small :-)

src/utils/Config.ts
```diff
@@ -0,0 +1,18 @@
+import * as dotenv from 'dotenv';
+
+export const loadConfig = () => {
```
geek-soft: I have very good experience with https://www.npmjs.com/package/config. It by default supports files per environment (in many formats including JSON and YAML) and it has support for environment variables as possible way to override defaults.

In terms of choice I guess it comes down to personal preference.

If you wanna stick with `dotenv` please simplify config loading. You can basically validate environment type using regex for example and then construct use the value to dynamically evaluate path e.g. `path = '${__dirname}/../../.env.${process.env.NODE_ENV}'`. Having some many switch cases will make maintenance more difficult.

src/middleware/SessionMiddleware.ts
geek-soft: Agree with @n-r-b

src/middleware/SessionMiddleware.ts
```diff
@@ -0,0 +1,29 @@
...
+    public handler(req: Request, res: Response, next: NextFunction): void {
+        this.redisService.getObject('session').then((session: any) => {
+            console.log(session)
+            if (!session) {
+                console.log('No session. Creating a new session');
+                this.redisService.setObject('session',  {example: 'test', data: 'cookie'});
+            } else {
+                console.log('Session exists');
+            }
+            console.log('next');
```
akaJSaunders: console.log('next') sounds very vague, a more complete message of what is happening would be more useful when debugging

src/services/RedisService.ts
```diff
@@ -0,0 +1,27 @@
+import { RedisClient } from 'redis';
+import { injectable, inject } from 'inversify';
+import { TYPES } from '../Types';
+import { IMap } from '../utils/Types';
+import { getAsync, setAsync, setObjectAsync, getObjectAsync } from '../utils/RedisUtils';
```
geek-soft: I suggest we move it into RedisService. It brings no value outside RedisService and for that reason should not be as public API.

src/Server.ts
```diff
@@ -35,6 +38,16 @@ export class Server {
 
   private createContainerWithBindings(): Container {
     const container = new Container();
+    container.bind<RedisService>(TYPES.RedisService).to(RedisService);
```
geek-soft: Since more and more things were added to the container I suggest you move it off the `Server.ts` cause in current shape it has too many responsibilities reasons to change. It definitely brakes Single Responsibility Principle.

src/utils/Config.ts
geek-soft: https://www.npmjs.com/package/dotenv#what-happens-to-environment-variables-that-were-already-set <- @n-r-b , @akaJSaunders worth to have a read if we stick to `dotenv`.

.env.example
```diff
@@ -0,0 +1,2 @@
+REDIS_HOST=<HOST_HERE>
```
geek-soft: I do not see much value in in. I'd rather have exemplar file with values for local environment or development environment, something which has no secret secrets.

src/services/RedisService.ts
franciscog-kainos: The point was to segregate the implementation of the promises away from the controller to make it more readable

src/Server.ts
franciscog-kainos: Agreed 👍

n-r-b: Instead of binding to the container in this way, we could use decorators: https://github.com/inversify/inversify-binding-decorators

src/middleware/SessionMiddleware.ts
franciscog-kainos: Agreed 👍

src/utils/RedisUtils.ts
franciscog-kainos: Oops that's a good point

src/App.ts
```diff
@@ -1,4 +1,12 @@
 import { Server } from './Server';
+import { loadConfig } from './utils/Config';
+
+/**
+ * Load environment config.
+ */
+const config = loadConfig();
+if (config.error)
+    console.error(config.error);
```
geek-soft: I am not sure what make `config.error` but I'd rather crash the app than progress with invalid configuration in the fail-fast spirit.

src/utils/Config.ts
franciscog-kainos: Let me know which library you prefer for this, I don't mind either but I've used dotenv before

src/Server.ts
franciscog-kainos: Yes that's much nicer thanks!

src/services/RedisService.ts
geek-soft: I do understand however turning callbacks into promises is delegated to helper method anyway making it one liner. I do not see much of an issue having `promisify(client.get).bind(client)` in implementation of service methods. In my mind it makes more sense than making helpers methods available for potential reuse.

src/utils/Config.ts
geek-soft: Lets stick to `dotenv` since you started. As long as there is easy way to configure service via environment variables I am happy.

.gitignore
```diff
@@ -3,4 +3,6 @@ node_modules/
 *.zip
 .idea/
 .DS_Store
-.vscode/
\ No newline at end of file
+.vscode/
+.env.*
+!.env.local
```
n-r-b: if we ignore this, then we are forced to create a local config file?

franciscog-kainos: This prevents the .env.local from being ignored

franciscog-kainos: But yes, if we do that then we will have to create one ourselves

src/controllers/AbstractRestController.ts
```diff
@@ -0,0 +1,29 @@
+import { inject } from 'inversify';
+import { BaseHttpController, httpGet, controller } from 'inversify-express-utils';
+import { RedisService } from '../services/RedisService';
+import { TYPES } from '../InjectableTypes';
+
+@controller('/')
+export abstract class AbstractRestController extends BaseHttpController {
```
n-r-b: The name 'AbstractRestController' doesn't really describe it's content

franciscog-kainos: Maybe `HealthCheckController`?

src/controllers/AbstractRestController.ts
```diff
@@ -0,0 +1,29 @@
...
+    public async healthCheck(): Promise<void> {
+        await this.sendInfo();
+    }
+
+    private async sendInfo(): Promise<void> {
+        const obj: any = await this.redisService.checkHealth();
+        this.httpContext.response.send(
+            `
+            <p>Redis online:  ${this.redisService.ping()}</p>
+            <p>Number of Times Visited: ${obj.n}\n</p>
```
n-r-b: Is it relevant know the no. times visited and last time visited? Do we just want to know the connection status?

franciscog-kainos: It shows evidence that we can retrieve and store data from redis which is quite important since we don't have a testing framework yet

src/middleware/ErrorHandler.ts
```diff
@@ -0,0 +1,8 @@
+import { Request, Response, NextFunction } from 'express';
+
+export function handler(err: any, req: Request, res: Response, nextFunction: NextFunction): void {
+    console.error(err.message);
+    if (!err.statusCode) err.statusCode = 500;
```
n-r-b: we could use a library for the status codes https://www.npmjs.com/package/http-status-codes

franciscog-kainos: I like that 👍

src/services/RedisService.ts
```diff
@@ -0,0 +1,136 @@
...
+import { TYPES } from '../InjectableTypes';
+import { IMap } from '../utils/Types';
+import { promisify } from 'util';
+import { createClient, RedisClient } from 'redis';
+
+@provide(TYPES.RedisService)
+export class RedisService {
+
+    private healthObject: any = {
+        n: 0,
```
n-r-b: what is 'n' here?

src/services/RedisService.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    public async checkHealth(): Promise<any> {
+        return this.getObject('healthCheck').then(obj => {
+            if (obj) {
+                this.healthObject = obj;
+                this.healthObject.n = Number(this.healthObject.n) + 1;
+                this.healthObject.lastTimeVisited = new Date().getTime().toString();
+                this.setObject('healthCheck', this.healthObject);
+            } else {
+                this.setObject('healthCheck', this.healthObject);
```
n-r-b: So if there is no connection to redis, then we will still return the lastTimeVisited?

src/utils/ConfigLoader.ts
```diff
@@ -0,0 +1,12 @@
+import * as dotenv from 'dotenv';
+
+const DEFAULT_ENV_FILE = `${__dirname}/../../.env.local`;
+
+export const loadConfig = () => {
+  if (process.env.NODE_ENV) {
+    const path = `${__dirname}/../../.env.${process.env.NODE_ENV}`;
+    return dotenv.config({ path });
+  }
+  return dotenv.config({ path: DEFAULT_ENV_FILE });
```
geek-soft: I suggest we do not load local file by default because it will load it in other non local environment.

As discussed I see this to he following way:
- load default `.env` file
- load `.env.[environment]` file if `NODE_ENV` is set

Beside above we should keep `.env.local` file for local environment but ask ourselves to do `NODE_ENV=local npm start` locally. Alternatively we could add it to start script in package.json (subject to future change when we figured out how to run it in cloud)

src/App.ts
```diff
@@ -1,4 +1,12 @@
 import { Server } from './Server';
+import { loadConfig } from './utils/ConfigLoader';
+
+/**
```
geek-soft: I do not find these comments useful. Code should be self explanatory.

src/services/RedisService.ts
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Function used as the fail function in this context.
+ */
+const logError = () => console.error('Redis Not online');
+/**
+ * Partially applied function which uses the above function as the fail function.
+ * @param cond The condiditon function
+ */
+const standardAction = (cond: () => boolean) => actOnConnectionStatus(cond)(logError);
```
n-r-b: do we need to declare a const for this? - when there's no additional logic

src/App.ts
geek-soft: We could rename function to `loadEnvironmentVariablesFromFiles()` - not so short but does not require comment IMO.

src/services/RedisService.ts
franciscog-kainos: image

franciscog-kainos: Yes

src/services/RedisService.ts
franciscog-kainos: Good point 👍

src/controllers/AbstractRestController.ts
n-r-b: Sounds good 👍

src/services/RedisService.ts
n-r-b: Could we format the date into something more readable?

src/controllers/AbstractRestController.ts
n-r-b: ok, that's fair enough for the time being

src/services/RedisService.ts
franciscog-kainos: It partially applies the logError function so that if we want to change how the errors get logged we only need to swap out that function for a different one

src/App.ts
```diff
@@ -1,4 +1,12 @@
...
+ * Load environment config.
+ */
+const config = loadConfig();
+if (config.error)
```
geek-soft: I suggest we always use brackets to avoid potential problems in the future.

geek-soft: Also from what I noticed it throw an error when file does not exist which is 👍 . I did not notice same error when file content wasn't in key-value pair format but that is fine.

An example of error:

```
Error: ENOENT: no such file or directory, open '/Users/ddunajski/Projects/lfp-appeals-frontend/build/utils/../../.env.fake'
    at Object.openSync (fs.js:457:3)
    at Object.readFileSync (fs.js:359:35)
    at Object.config (/Users/ddunajski/Projects/lfp-appeals-frontend/node_modules/dotenv/lib/main.js:96:29)
    at Object.exports.loadConfig (/Users/ddunajski/Projects/lfp-appeals-frontend/build/utils/ConfigLoader.js:8:23)
    at Object.<anonymous> (/Users/ddunajski/Projects/lfp-appeals-frontend/build/App.js:5:31)
    at Module._compile (internal/modules/cjs/loader.js:1151:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/Users/ddunajski/Projects/lfp-appeals-frontend/build/utils/../../.env.fake'
}
```

It looks readable to me so 👍

src/controllers/AbstractRestController.ts
```diff
@@ -0,0 +1,29 @@
+import { inject } from 'inversify';
+import { BaseHttpController, httpGet, controller } from 'inversify-express-utils';
+import { RedisService } from '../services/RedisService';
+import { TYPES } from '../InjectableTypes';
+
+@controller('/')
+export abstract class AbstractRestController extends BaseHttpController {
+
+    constructor(@inject(TYPES.RedisService) protected readonly redisService: RedisService) {
```
geek-soft: It could be made private. No need to have it protected at this point.

src/controllers/AbstractRestController.ts
geek-soft: Health check should communicate correctness through status code so observers does not have to parse response. It usually does 200 for OK and 500 for not OK.

src/services/RedisService.ts
geek-soft: As far as service is concerned IMO it should just say healthy / unhealthy, nothing more.

src/utils/ConfigLoader.ts
```diff
@@ -0,0 +1,20 @@
+import * as dotenv from 'dotenv';
+
+const DEFAULT_ENV_FILE = `${__dirname}/../../.env`;
+
+const loadConfigHelper = () => {
```
geek-soft: I'd change that name as I am not sure what helper means in that context. Do you mean that config object returned by `dotenv.config` is a helper? Or is that method a helper method?

src/controllers/LandingController.ts
```diff
@@ -1,11 +1,12 @@
...
 import { controller, httpGet } from 'inversify-express-utils';
+import { OK } from 'http-status-codes';
 
 @controller('/')
 export class LandingController {
 
     @httpGet('')
     public renderView(req: Request, res: Response, next: NextFunction): void {
-        res.render('landing');
+        res.status(OK).render('landing');
```
geek-soft: I am not sure we need that, 200 is used by default.

src/utils/ConfigLoader.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (process.env.NODE_ENV) {
```
geek-soft: It is still not quite right. Default config `.env` should be loaded at all times. Next environment specific config is loaded to override default / shared config if needed and add extra configuration. Something like:

```
dotenv.config();
if (process.env.NODE_ENV) {
  dotenv.config(`${__dirname}/../../.env.${process.env.NODE_ENV}`)
}
```

franciscog-kainos: I see

PR #11, opened 2020-02-04
build: added watch ability to project

franciscog-kainos:
This pull request added the ability to run the project locally with auto-reloading after changes are made.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "scripts": {
     "start": "npm run build && NODE_ENV=local node build/App.js",
     "build": "gulp build",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "start": "npm run build && NODE_ENV=local node build/App.js",
     "build": "gulp build",
+    "watch": "npm run build:clean && gulp start:watch",
```
geek-soft: Wouldn't `build` instead of `build:clean` be enough?

franciscog-kainos: Just for good measure I suppose

PR #15, opened 2020-02-07
feat: add screen capturing other appeal reason

geek-soft:
### JIRA link

TBC

### Change description

Adds screen capturing other appeal reason.

image

image

### Work checklist

### Merge instructions

```
```


src/controllers/OtherReasonController.ts
```diff
@@ -1,14 +1,48 @@
...
     }
 
-    @httpGet(OTHER_REASON_ENTRY_PREFIX)
-    public showDisclaimer(): void {
-        this.httpContext.response.render('other-reason-disclaimer');
+    @httpGet(OTHER_REASON_PAGE_PREFIX)
+    public async renderForm(): Promise<void> {
+        const session = await this.redisService.get(sessionKey);
+
+        return this.render(session ? JSON.parse(session) : {})
```
n-r-b: I see we are rendering the view with the 'session', where does this get set?

src/models/OtherReason.schema.ts
```diff
@@ -0,0 +1,25 @@
+import * as Joi from '@hapi/joi';
+
+const titleErrorMessage = 'You must give your reason a title';
+const descriptionErrorMessage = 'You must give us more information';
+
+export const schema = Joi.object({
+    title: Joi.string()
+        .required()
+        .pattern(/\w+/)
```
akaJSaunders: I dont think that regex is necessary, as in the message 'string.empty' already captures cases where nothing is input

src/controllers/OtherReasonController.ts
```diff
@@ -1,14 +1,48 @@
...
 export class OtherReasonController extends BaseHttpController {
-    constructor() {
+    constructor(@inject(RedisService) private readonly redisService: RedisService) {
         super();
     }
 
-    @httpGet(OTHER_REASON_ENTRY_PREFIX)
-    public showDisclaimer(): void {
-        this.httpContext.response.render('other-reason-disclaimer');
+    @httpGet(OTHER_REASON_PAGE_PREFIX)
```
n-r-b: Since we are duplicating the 'OTHER_REASON_PAGE_PREFIX' across two endpoints, is it worth putting this at the top at controller level?

src/models/OtherReason.schema.ts
```diff
@@ -0,0 +1,25 @@
+import * as Joi from '@hapi/joi';
+
+const titleErrorMessage = 'You must give your reason a title';
+const descriptionErrorMessage = 'You must give us more information';
+
+export const schema = Joi.object({
+    title: Joi.string()
+        .required()
```
akaJSaunders: Is the required() check necessary? I would assume string.empty message would handle the case of an empty field

src/controllers/OtherReasonController.ts
```diff
@@ -1,14 +1,48 @@
...
+    public async handleFormSubmission(): Promise<void> {
+        const body: OtherReason = this.httpContext.request.body;
+
+        const validationResult: ValidationResult = new SchemaValidator(schema).validate(body);
+
+        if (validationResult.errors.length === 0) {
+            await this.redisService.set(sessionKey, JSON.stringify(body))
+        }
+
+        return this.render({...body, validationResult});
```
n-r-b: If there is a validation error, would it make sense to set the status code to reflect this? i.e. 400 Bad Request

n-r-b: Also, are we parsing all responses as json?

src/controllers/OtherReasonController.ts
```diff
@@ -1,14 +1,48 @@
...
+    }
+
+    private async render(data: any): Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            this.httpContext.response.render('other-reason', data, (err, compiled) => {
```
n-r-b: how do you feel about constants for the template name?

test/controllers/OtherReasonController.test.ts
```diff
@@ -1,18 +1,71 @@
...
+                .expect(response => {
+                    expect(response.status).to.be.equal(200);
+                    expect(response.text).to.include(pageHeading)
+                        .and.not.include(errorSummaryHeading);
+                });
+        });
+    });
+
+    describe('POST request', () => {
+        it('should return 200 response with rendered error messages when invalid data was submitted', async () => {
```
n-r-b: Again, is it worth return 400 here?

src/models/OtherReason.schema.ts
```diff
@@ -0,0 +1,25 @@
...
+    description: Joi.string()
+        .required()
+        .pattern(/\w+/)
+        .messages({
+            'any.required': descriptionErrorMessage,
+            'string.base': descriptionErrorMessage,
+            'string.empty': descriptionErrorMessage,
+            'string.pattern.base': descriptionErrorMessage
+        })
+});
```
akaJSaunders: Couldn't we have a unified schema.ts with multiple schema constants for different pages, rather than having a separate file for each one

src/controllers/OtherReasonController.ts
geek-soft: No

test/controllers/OtherReasonController.test.ts
geek-soft: Changed to 422.

PR #16, opened 2020-02-07
feat: Other reason disclaimer screen

franciscog-kainos:
Also refactored code to extract express app config so it can be used in tests.

### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1272

### Change description
Added a new controller to handle the new route for the "Other Reason Disclaimer Screen".
Extracted the config of the express app into a function so it can be reused in the tests.
Created simple unit test.

### Work checklist

- [ * ] Tests added where applicable
- [ * ] UI changes look good on mobile
- [ * ] UI changes meet accessibility criteria

### Merge instructions

```
```

### DIFF
--- a/src/public/sass/application.scss
+++ b/src/public/sass/application.scss
@@ -3,4 +3,5 @@ $govuk-global-styles: true;
 @import 'govuk-frontend/govuk/all';
 
 // Patterns that aren't in govuk-frontend
-@import "patterns/related-items";
+@import 'patterns/related-items';
+@import 'other-reason/extra-styles'
--- a/src/public/sass/patterns/_related-items.scss
+++ b/src/public/sass/patterns/_related-items.scss
@@ -2,11 +2,10 @@
 // can be seen at https://components.publishing.service.gov.uk/component-guide/related_navigation
 
 .app-related-items {
-    border-top: 10px solid govuk-colour("blue");
-    padding-top: govuk-spacing(2);
-  }
-  
-  .app-related-items .govuk-list > li {
-    margin-bottom: govuk-spacing(2);
-  }
-  
+  border-top: 10px solid govuk-colour("blue");
+  padding-top: govuk-spacing(2);
+}
+
+.app-related-items .govuk-list > li {
+  margin-bottom: govuk-spacing(2);
+}
--- a/src/views/other-reason-disclaimer.njk
+++ b/src/views/other-reason-disclaimer.njk
@@ -1,0 +1,52 @@
+{% extends "layout.njk" %}
+
+{% from "govuk/components/button/macro.njk" import govukButton %}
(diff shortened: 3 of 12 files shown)


src/public/sass/application.scss
```diff
@@ -1,6 +1,89 @@
...
+  }
+
+  .circle-white {
+    background-color: govuk-colour("white");
+    color: govuk-colour("black");
+    margin-right: 0.5em;
+    position: relative;
+    top: -0.3em;
+  }
+}
```
akaJSaunders: I was wondering if maybe we should store these in a separate scss file under the patterns directory and then in the applications.scss we can just import them? Curious what you think @franciscog-kainos

src/views/other-reason-disclaimer.njk
```diff
@@ -0,0 +1,52 @@
...
+            text: "Continue with appeal",
+            href: "#",
+            classes: "govuk-link govuk-!-margin-top-5 govuk-link--interrupt"
+          }) 
+        }}
+        <p>
+          <a href="#" class="govuk-link">Save and sign out</a>
+        </p>
+  </div>
+    {% endblock %}
```
akaJSaunders: should probably remove unecessary indentation

src/public/sass/application.scss
```diff
@@ -1,6 +1,89 @@
 $govuk-global-styles: true;
 
-@import 'govuk-frontend/govuk/all';
+@import "govuk-frontend/govuk/all";
```
akaJSaunders: Not sure what the convention is, but my VScode keeps complaining when I use " " with yellow squiggly lines, and insists I use ' '.  I realise this is a rather small thing, but what do you think? (

franciscog-kainos: Yep should be single quotes 👍

src/public/sass/application.scss
franciscog-kainos: Good point

src/public/sass/patterns/_related-items.scss
```diff
@@ -2,11 +2,93 @@
...
+  .circle-white {
+    background-color: govuk-colour("white");
+    color: govuk-colour("black");
+    margin-right: 0.5em;
+    position: relative;
+    top: -0.3em;
   }
-  
\ No newline at end of file
+}
```
akaJSaunders: this file was called _related-items because the original contents were part of the related-items section (or as we were calling it subsection) of the landing page. Im not too bothered about how we distribute things, but we should probably rename this file if we are going to use it for multiple components

franciscog-kainos: Check if you like it like this now

akaJSaunders: Looks good! Thanks mate!

PR #22, opened 2020-02-12
build: add makefile and start scripts

Krzysztofoc:
### Change description

Adding Makefile, start.sh, version and routes.yml to make pipeline working properly with already existing pipeline jobs definitions.

### Work checklist

### Merge instructions

```
```


version
```diff
@@ -0,0 +1 @@
+0.1
```
geek-soft: How is that file used?

routes.yaml
```diff
@@ -0,0 +1,6 @@
+app_name: lfp-appeals-frontend
+group: web
+weight: 900
+routes:
+  1: /appeal-a-penalty
```
geek-soft: Where did you take it from?

Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+.PHONY: clean
+clean:
+	rm -f ./$(artifact_name)-*.zip
+	rm -rf ./build
+	rm -rf ./temp-*
+	rm -f ./build.log
+
+package-install:
+	npm i
```
geek-soft: I suggest we use full command name e.g. `install` in our case.

routes.yaml
Krzysztofoc: 5th Feb conversation on the lfp-appeals slack channel

version
Krzysztofoc: It's used by one of the pipeline steps and then it's passed to application versioning and GitHub tags. 
Every rebuild will add +1:
0.1.1
0.1.2 etc.

Makefile
```diff
@@ -0,0 +1,52 @@
...
+	npm run lint
+
+.PHONY: sonar
+sonar:
+#	npm run sonarqube
+
+.PHONY: test
+test: test-unit
+
+.PHONY: test-unit
```
n-r-b: what's the reason for running sonarqube before the tests?

PR #23, opened 2020-02-12
chore: add schema validator

geek-soft:
### JIRA link

None

### Change description

Add schema validator

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "test": "mocha test/**/*.test.ts"
   },
   "dependencies": {
+    "@hapi/joi": "^17.1.0",
     "body-parser": "^1.18.2",
     "cookie-parser": "^1.4.4",
     "dotenv": "^8.2.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -40,6 +41,7 @@
     "@types/dotenv": "^8.2.0",
     "@types/express": "^4.0.37",
     "@types/express-session": "^1.15.3",
+    "@types/hapi__joi": "^16.0.9",
```
n-r-b: i don't think this one is being used?

package.json
```diff
@@ -18,6 +18,7 @@
     "test": "mocha test/**/*.test.ts"
   },
   "dependencies": {
+    "@hapi/joi": "^17.1.0",
```
n-r-b: Should this be a developer dependency?

geek-soft: No, validation is needed at runtime. Anything that application needs to do the job should be defined as regular dependency.

package.json
geek-soft: It is, the other one is not.

PR #24, opened 2020-02-12
refactor: change redis creation to use url

franciscog-kainos:
### JIRA link
N/A

### Change description
Changed use of redis port and host to use the full url.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/ContainerFactory.ts
+++ b/src/ContainerFactory.ts
@@ -3,12 +3,13 @@ import { RedisClient, createClient } from 'redis';
 import { buildProviderModule } from 'inversify-binding-decorators';
 
 const createRedisClient = () => {
-    return createClient(
-        {
-            host: process.env.REDIS_HOST,
-            port: Number(process.env.REDIS_HOST)
-        }
-    ).on('error', (err) => { throw err; });
+
+    if (process.env.CACHE_SERVER) {
+        throw Error('CACHE_SERVER variable not set.');
+    }
+    const redisUrl = `redis://${process.env.CACHE_SERVER}`;
+
+    return createClient(redisUrl).on('error', (err) => { throw err; });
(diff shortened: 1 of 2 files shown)


src/ContainerFactory.ts
```diff
@@ -3,12 +3,8 @@ import { RedisClient, createClient } from 'redis';
 import { buildProviderModule } from 'inversify-binding-decorators';
 
 const createRedisClient = () => {
-    return createClient(
-        {
-            host: process.env.REDIS_HOST,
-            port: Number(process.env.REDIS_HOST)
-        }
-    ).on('error', (err) => { throw err; });
+    const redisUrl = 'redis://' + process.env.CACHE_SERVER || 'localhost:6379';
```
geek-soft: No default please. We expect to fail is config is missing.

geek-soft: Not a big deal but why not to use string interpolation instead of '+'?

franciscog-kainos: Good point, fixed those

src/ContainerFactory.ts
```diff
@@ -3,12 +3,8 @@ import { RedisClient, createClient } from 'redis';
 import { buildProviderModule } from 'inversify-binding-decorators';
 
 const createRedisClient = () => {
-    return createClient(
-        {
-            host: process.env.REDIS_HOST,
-            port: Number(process.env.REDIS_HOST)
-        }
-    ).on('error', (err) => { throw err; });
+    const redisUrl = `redis://${process.env.CACHE_SERVER}`;
```
geek-soft: Ideally if you checked whether `process.env.CACHE_SERVER` is set and throw readable error when not set.

PR #31, opened 2020-02-14
feat: link between disclaimer and other reason details

n-r-b:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1300

### Change description
Transition between disclaimer and other reason details pages implemented with a POST endpoint on the Disclaimer Controller.

### Work checklist

### Merge instructions

```
```


src/views/other-reason-disclaimer.njk
```diff
@@ -38,15 +38,16 @@
...
-        </p>
+        <form method="post">
+          {{ 
+            govukButton({
+              text: "Continue with appeal",
+              classes: "govuk-link govuk-!-margin-top-5 govuk-link--interrupt"
+            }) 
+          }}
+          <p>
+            <a href="#" class="govuk-link">Save and sign out</a>
```
mario-paolucci: we should probably get rid of this button and not include it on any other page even though the prototype has it, since it's not something that's going to be implemented in our first release

n-r-b: we'll have to run this past our lovely BA/QA

PR #40, opened 2020-02-17
Lfa1253/sign on page

franciscog-kainos:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1260

### Change description
Added `ch-session-node` api which is currently under review [PR for poc-session-handler](https://github.com/companieshouse/poc-node-session-handler/pull/2) 

### Work checklist

### Merge instructions

```
```


src/views/landing.njk
```diff
@@ -32,7 +32,8 @@
             isStartButton: true,
             attributes: {
               id: "submit"
-            }
+            },
+            href: penaltyDetailsPage
```
akaJSaunders: I think we were avoiding adding transitions to other pages through the template files. Currently in master, landing page redirects through a post request (when button is clicked) to a entry controller ('/start') which calls a get request that redirects to penalty details page. This is because landing page is ultimately a throwaway page. Is this explanation correct @geek-soft ? haha

src/App.ts
```diff
@@ -1,7 +1,7 @@
 import { Server } from './Server';
-import { loadEnvironmentVariablesFromFiles } from './utils/ConfigLoader';
+import { loadEnvironmentVariablesFromFiles, ENV } from './utils/ConfigLoader';
 
 loadEnvironmentVariablesFromFiles();
 
-const server = new Server(Number(process.env.PORT) || 3000);
+const server = new Server(ENV.ERIC_PORT);
```
geek-soft: How is that related to the user story?

src/views/landing.njk
geek-soft: Redirect is done on submission, you are right @akaJSaunders it is not needed.

src/utils/IConfig.ts
```diff
@@ -0,0 +1,10 @@
+export default interface IConfig {
+    readonly ERIC_PORT: number;
+    readonly CACHE_SERVER: string;
+    readonly CACHE_DB: number;
+    readonly CACHE_PASSWORD: string;
+    readonly COOKIE_NAME: string;
+    readonly COOKIE_SECRET: string;
+    readonly DEFAULT_SESSION_EXPIRATION: number;
```
geek-soft: How is that used or intended to be used?

src/utils/ConfigLoader.ts
```diff
@@ -14,11 +15,14 @@ const checkFileExists = (config: dotenv.DotenvConfigOutput) => {
...
 };
 
+export let ENV: IConfig;
+
 export const loadEnvironmentVariablesFromFiles = () => {
   dotenv.config({ path: DEFAULT_ENV_FILE });
   if (process.env.NODE_ENV) {
     const envFilePath = `${__dirname}/../../.env.${process.env.NODE_ENV}`;
     checkFileExists(dotenv.config({ path: envFilePath }));
+    ENV = new Config();
```
geek-soft: It looks like surprising side effect. I guess implementation might not be complete but at this point doing validation in constructor does not look like the right thing.

package.json
```diff
@@ -29,6 +30,7 @@
     "inversify": "^5.0.1",
     "inversify-binding-decorators": "^4.0.0",
     "inversify-express-utils": "^6.3.2",
+    "msgpack5": "^4.2.1",
```
geek-soft: Removal candidate, `ch-node-session` should bring it as transitive dep.

src/utils/IConfig.ts
franciscog-kainos: the process.env give out variables of type `T  | undefined`. The point was centralise the checking of ENV variables so that we don't have to check every time if process.env variable is undefined

src/middleware/AuthMiddleware.ts
```diff
@@ -0,0 +1,22 @@
...
+export class AuthMiddleware extends BaseMiddleware {
+
+    public constructor() {
+        super();
+    }
+
+    public handler: RequestHandler = expressAsyncHandler((req: Request, res: Response, next: NextFunction): void => {
+
+        if (req.session.isNothing()) {
+            res.redirect(`http://localhost:3001/signin?return_to=${req.url}`);
```
n-r-b: Can we create a constant for this path and make the env configurable?

src/middleware/AuthMiddleware.ts
```diff
@@ -0,0 +1,22 @@
...
+        }
+        console.log('Authenticated');
```
n-r-b: How do we know at this point the sign in was successful?  Or do we assume it was and then have a check in the controller?

franciscog-kainos: Check new version

src/controllers/OtherReasonController.ts
```diff
@@ -5,33 +5,49 @@ import { SchemaValidator } from '../utils/validation/SchemaValidator';
...
         super();
     }
 
     @httpGet('')
-    public async renderForm(): Promise<void> {
-        const session = await this.redisService.get(sessionKey);
+    public async renderForm(req: Request): Promise<void> {
+
+        const data = req.session
+            .chain(session => session.getExtraData())
```
n-r-b: Could the middleware handle this?

src/controllers/OtherReasonController.ts
```diff
@@ -5,33 +5,49 @@ import { SchemaValidator } from '../utils/validation/SchemaValidator';
...
+            .map(extraData => extraData[sessionKey]);
+
+        if (data.isJust()) {
```
n-r-b: we could simplify this: `return await this.render(OK, data.isJust() ? data.__value: {});`

src/controllers/OtherReasonController.ts
```diff
@@ -5,33 +5,49 @@ import { SchemaValidator } from '../utils/validation/SchemaValidator';
...
+import { EitherUtils, SessionMiddleware, SessionStore } from 'ch-node-session';
+import { AuthMiddleware } from '../middleware/AuthMiddleware';
 
 const sessionKey = 'session::other-reason';
 
-@controller(OTHER_REASON_PAGE_URI)
+@controller(OTHER_REASON_PAGE_URI, SessionMiddleware, AuthMiddleware)
 export class OtherReasonController extends BaseHttpController {
-    constructor(@inject(RedisService) private readonly redisService: RedisService) {
+    constructor(@inject(SessionStore) private readonly store: SessionStore) {
```
n-r-b: can we call this `sessionStore`? as there is a `store` method on the SessionStore class

src/controllers/OtherReasonController.ts
```diff
@@ -5,33 +5,49 @@ import { SchemaValidator } from '../utils/validation/SchemaValidator';
...
+        const body: OtherReason = req.body;
 
         const validationResult: ValidationResult = new SchemaValidator(schema).validate(body);
         const valid = validationResult.errors.length === 0;
 
         if (valid) {
-            await this.redisService.set(sessionKey, JSON.stringify(body))
+            req.session.map(async session => {
+                session.saveExtraData(sessionKey, body);
+                EitherUtils.wrapEitherFunction(Cookie.validateCookieString(req.cookies.__SID));
```
n-r-b: Are we not validating the cookie in the session middleware?

src/controllers/OtherReasonController.ts
franciscog-kainos: What do you mean?

src/middleware/AuthMiddleware.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    public handler: RequestHandler = (req: Request, res: Response, next: NextFunction): void => {
+
+        const returnURL = this.popReturnURL();
+
+        if (req.cookies.__SID) {
+            const signInInfo = req.session.chain(session => session.getValue<ISignInInfo>(SessionKey.SignInInfo));
+            signInInfo.map(info => {
+                if (info[SignInInfoKeys.SignedIn] !== 1) {
+                    res.redirect(`${returnEnvVarible('REDIRECT_URL')}?returnTo=${returnURL}`);
```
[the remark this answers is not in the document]
geek-soft: Query param should be named `return_to` not `returnTo` otherwise redirect does not happen.

src/controllers/EntryController.ts
```diff
@@ -1,11 +1,19 @@
...
 
+    constructor(@inject(AuthMiddleware) private readonly auth: AuthMiddleware) {
+        super();
+    }
+
     @httpGet('')
-    public redirectView(): void {
-        this.httpContext.response.redirect(PENALTY_DETAILS_PAGE_URI);
+    public async redirectView(): Promise<any> {
+        this.auth.setReturnURL(PENALTY_DETAILS_PAGE_URI);
```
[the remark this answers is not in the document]
franciscog-kainos: image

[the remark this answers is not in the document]
franciscog-kainos: The acceptance criteria knows best here

package.json
```diff
@@ -43,6 +45,7 @@
     "@types/express": "^4.0.37",
     "@types/express-session": "^1.15.3",
     "@types/hapi__joi": "^16.0.10",
+    "@types/ioredis": "^4.14.7",
```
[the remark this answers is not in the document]
franciscog-kainos: After doing clean-install i get this when i run it
image

src/controllers/LandingController.ts
```diff
@@ -1,17 +1,18 @@
...
+import { BaseAsyncHttpController } from './BaseAsyncHttpController';
 
 @controller(ROOT_URI)
-export class LandingController extends BaseHttpController {
+export class LandingController extends BaseAsyncHttpController {
 
     @httpGet('')
-    public renderView(req: Request, res: Response, next: NextFunction): void {
-        res.render('landing');
+    public async renderView(): Promise<string> {
```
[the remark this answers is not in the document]
franciscog-kainos: I'm testing these one by one to see which ones cause that dreadful `headers already sent to client `error

src/controllers/EntryController.ts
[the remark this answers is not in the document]
geek-soft: As far as I understand the story (and other stories in either sprint or backlog) it should always take to penalty details screen after sign in. Please make it hardcoded. There is no need for such complexity. Let's keep it simple.

src/controllers/ConfirmationController.ts
```diff
@@ -1,19 +1,19 @@
...
 
     @httpGet('')
-    public getConfirmationView(): void {
+    public getConfirmationView(req: Request): void {
 
-        const session = this.httpContext.request.session;
-        let companyNumber: string = '';
+        req.session.chain(_ => _.getExtraData())
+            .map(data => data.appeals.penaltyIdentifier.companyNumber)
+            .map(companyNumber => this.httpContext.response.render('confirmation', { companyNumber }))
```
geek-soft: It would be ideal if we did not have to make two render calls and structure code in a way so that it reads session first and then it renders.

test/ApplicationFactory.ts
```diff
@@ -1,10 +1,67 @@
...
+
+    const sessionStore = Substitute.for<SessionStore>();
+
+    if (session && cookie) {
+        sessionStore.load(cookie).returns(EitherUtils.wrapValue(session.data));
+    }
+
+    const realMiddleware = SessionMiddleware(config, sessionStore);
+    const sessionHandler = (req: Request, res: Response, next: NextFunction) => {
+        req.session = session ? Maybe.of(session) : Maybe.empty();
```
geek-soft: That seems to be not needed since real middleware will take care of it. Please remove.

test/controllers/OtherReasonController.test.ts
```diff
@@ -51,11 +56,12 @@ describe('OtherReasonController', () => {
             const title = 'Some title';
             const description = 'Some description';
 
-            const app = createApplication(container => {
-                container.bind(RedisService).toConstantValue(createSubstituteOf<RedisService>(service => {
-                    service.set('session::other-reason', JSON.stringify({ title, description })).returns(Promise.resolve())
-                }))
-            });
+            const sessionStore = Substitute.for<SessionStore>();
```
geek-soft: Please remove and put `{ title, description }` into session that is being passed to `createApp(session)` call.

test/controllers/OtherReasonDisclaimerController.test.ts
```diff
@@ -1,29 +1,48 @@
...
 import { OK, MOVED_TEMPORARILY } from 'http-status-codes';
+import { CookieConfig, SessionStore, SessionMiddleware } from 'ch-node-session-handler';
+import { Redis } from 'ioredis';
+import { RequestHandler } from 'express';
+import { getEnvOrDefault } from '../../src/utils/EnvironmentUtils';
+
 
 describe('OtherReasonDisclaimerController', () => {
 
+    const app = createAppConfigurable(container => {
```
geek-soft: Use the following combo here:
```
const session = createFakeSession([], config.cookieSecret, true);
const app = createApp(session);
```

:)

test/controllers/PenaltyDetailsController.test.ts
```diff
@@ -1,55 +1,44 @@
...
-        it('should return 200 when trying to access page without a session', async () => {
-            await request(app).get(PENALTY_DETAILS_PAGE_URI)
-                .expect(response => {
-                    expect(response.status).to.be.equal(OK);
-                    expect(response.text).to.contain(pageHeading)
-                        .and.not.contain(errorSummaryHeading);
-                })
+        it('should return 302 when trying to access page without a session', async () => {
+            await request(createApp()).get(PENALTY_DETAILS_PAGE_URI)
+                .expect(MOVED_TEMPORARILY);
```
geek-soft: This might not be enough. Check where user agent was moved to.

geek-soft: Same sort of test should be applied to other controllers protected by auth middleware. Could be next PR if we decided to keep this PR small. If so this test could be moved to separate PR.

test/middleware/AuthMiddleware.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        let penaltyDetails: Record<string, any>;
+
+        before('sign in server should be running', async () => {
+
+            penaltyDetails = {
+                penaltyReference: 'A12345678',
+                companyNumber: 'SC123123'
+            };
+
+            await request(createApp())
```
geek-soft: I think there is no point having this precondition here. Application that is created and tested here is thrown away and not used in any of the tests executed afterwards.

test/middleware/AuthMiddleware.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        it('should not redirect the user to the sign in page if the user is signed in', async () => {
+
+            for (const page of protectedPages) {
+                await request(authedApp).get(page)
+                    .expect(200);
+            }
+
+        });
+
+        it('should allow the user to go to the penalty reference screen if authed', async () => {
```
geek-soft: This is essentially same as previous test case. I suggest you remove it.

test/middleware/AuthMiddleware.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+            await request(authedApp)
+                .get(PENALTY_DETAILS_PAGE_URI)
+                .expect(200);
+
+        });
+        it('should continue to the next page without requiring auth if penalty details are valid', async () => {
```
geek-soft: There are few things I noticed about this test:
1) it targets penalty details page (integration) and it's behaviour while it should more focus on auth logic
2) it says 'if penalty details are valid' while validity of data does not matter at all since middleware is called before even controller gets the request and for that reason it is pointless

To fix it I suggest you change it to iterate through all the routes making POST request and assess that there is no redirect to sign in page.

Even better solution would be making this test a unit tests not integration tests but we have little time to change it.

test/middleware/AuthMiddleware.test.ts
```diff
@@ -0,0 +1,127 @@
...
+const sig = generateSignature(id, config.cookieSecret);
+
+const session = createFakeSession(
+    [{
+        [SessionKey.Id]: id
+    }, {
+        [SessionKey.ClientSig]: sig
+    }], config.cookieSecret, true);
+
+const authedApp = createApp(session);
```
geek-soft: Since it is only used in first describe block I suggest you move it inside the block

test/middleware/AuthMiddleware.test.ts
```diff
@@ -0,0 +1,127 @@
...
+            await request(authedApp)
+                .post(PENALTY_DETAILS_PAGE_URI)
+                .send(penaltyDetails)
+                .expect(_ => expect(_.header.location).to.include(OTHER_REASON_DISCLAIMER_PAGE_URI));
+
+        });
+    });
+    describe('Unauthed path', () => {
+
+        const appWithoutCookie = { label: 'App without Cookie', app: createApp(session, false) };
```
geek-soft: Label is not needed so please remove it. Variable name should be enough to express what it is.

geek-soft: Two things:
1) Auth middleware is not concerned with cookie.
2)` setCookie` in `const createApp = (session?: Session, setCookie: boolean = true)` is never used - likely because of 1)

I suggest you remove it due to above reasons.

test/middleware/AuthMiddleware.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        const unAuthedApp = { label: 'App without Cookie and Session', app: createApp() };
+
+        const unAuthedAppStateVariations = [appWithoutCookie, unAuthedApp];
+
+        before('sign in server should be running', async () => {
```
geek-soft: Same here, no value added, remove.

test/middleware/AuthMiddleware.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+        });
+
+    });
+
+    describe('Auth Edge case', () => {
+
+        const appWithoutSessionObject = createApp(createFakeSession([], config.cookieSecret), true);
+
+        before('sign in server should be running', async () => {
```
geek-soft: Same here, no value added, remove.

test/middleware/AuthMiddleware.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+            await request(createApp())
+                .get(PENALTY_DETAILS_PAGE_URI)
+                .catch(err => assert.fail(err));
+        });
+
+        it('should allow user to access protected pages if session object is somehow lost.', async () => {
```
geek-soft: Not sure what you trying to do here. What did you mean by 'is somehow lost'? A `createApp(createFakeSession([], config.cookieSecret), true)` call prepares session as other tests.

src/controllers/CheckYourAppealController.ts
```diff
@@ -1,24 +1,36 @@
...
+                    reasons: data.appeals.reasons
+                };
+            });
 
-        if(session){
-            penaltyIdentifier = session.getExtraData('appeals').penaltyIdentifier;
-            reasons = session.getExtraData('appeals').reasons;
-            userProfile = session.getValue('signin_info').user_profile;
-        }
+        userProfile
```
geek-soft: Mapping user profile and then getting appeals details does not make sense. Make just simple call to `this.httpContext.response.render` function passing both userProfile and appealsData. That mapping is absolutely necessary.

test/controllers/CheckYourAppealController.test.ts
```diff
@@ -1,27 +1,25 @@
...
-const app = createApplication(container => {
-    container.bind(RedisService).toConstantValue(createSubstituteOf<RedisService>());
-});
+const config = getDefaultConfig();
 
 describe('CheckYourAppealController', () => {
     describe('GET request', () => {
-        it('should return 200 when trying to access the submission summary', async () => {
-
+        it('should return 200 when trying to access the submission summary with a session ', async () => {
```
geek-soft: Not sure how much value that tests add. Not a concern of this PR but subject to removal in the future.

test/controllers/PenaltyDetailsController.test.ts
akaJSaunders: We have removed this test and in another PR we can address more tests regarding auth middleware

PR #60, opened 2020-02-26
feat: transit from summary page to confirmation page

geek-soft:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1286

### Work checklist

### Merge instructions

```
```


src/ContainerFactory.ts
```diff
@@ -1,10 +1,11 @@
 import { Container } from 'inversify';
 import { buildProviderModule } from 'inversify-binding-decorators';
 import { CookieConfig, SessionMiddleware, SessionStore } from 'ch-node-session-handler';
-import IORedis = require('ioredis');
-import { RequestHandler } from 'express';
 import { getEnvOrDefault } from './utils/EnvironmentUtils';
 import { AuthMiddleware } from './middleware/AuthMiddleware';
+import * as IORedis from 'ioredis'
+import * as kafka from 'kafka-node'
+import { EmailService } from './modules/email-publisher/EmailService'
```
geek-soft: The intention is to extract `modules/email-publisher` into separate repository as an external NPM module.

src/controllers/CheckYourAppealController.ts
```diff
@@ -26,7 +34,38 @@ export class CheckYourAppealController extends BaseAsyncHttpController {
...
+                    templateName: 'lfp-appeal-submission-confirmation',
+                    templateData: {
+                        companyNumber: appealsData[AppealKeys.PENALTY_IDENTIFIER][PenaltyIdentifierKeys.COMPANY_NUMBER],
+                        userProfile: {
+                            email: userProfile.email
+                        }
+                    }
+                }
+            })
+        } catch (err) {
```
geek-soft: I need to double check it with Bart. It feels like confirmation email is less critical and for that reason I decided to render no error message when sending failed.

src/controllers/CheckYourAppealController.ts
```diff
@@ -26,7 +34,38 @@ export class CheckYourAppealController extends BaseAsyncHttpController {
...
+            .extract() as Appeal;
 
+        try {
+            await this.emailService.send({
+                to: userProfile.email as string,
+                subject: 'Your appeal has been submitted',
+                body: {
+                    templateName: 'lfp-appeal-submission-confirmation',
+                    templateData: {
+                        companyNumber: appealsData[AppealKeys.PENALTY_IDENTIFIER][PenaltyIdentifierKeys.COMPANY_NUMBER],
```
akaJSaunders: I believe in a past PR, you commented on the excessive use of keys? Not entirely sure what you said, but are we going to continue to actively use these keys? Isn't there a better way of doing it? We can probably look into it in another PR but im just mainly curious if there is a neater way of doing things.

src/controllers/CheckYourAppealController.ts
akaJSaunders: From a developers point of view, isn't it handy for us to throw a proper error so that we can essentially fail early if something goes wrong?

test/ApplicationFactory.ts
```diff
@@ -24,7 +24,7 @@ export const getDefaultConfig = () => {
     };
 };
 
-export const createApp = (session?: Session) => createAppConfigurable(container => {
+export const createApp = (session?: Session, configureContainerBindings: (container: Container) => void = () => { }) => createAppConfigurable(container => {
```
akaJSaunders: This line feels a bit long, could we split into two lines for readability sake

src/views/confirmation.njk
```diff
@@ -15,6 +15,12 @@
         })
       }}
 
+      <p class="govuk-body-l">
+        We’ve sent a confirmation email to
```
akaJSaunders: What if we cant send an email to the given address?

src/ContainerFactory.ts
akaJSaunders: Is it going to be used by other services other than ours as far as we know?

src/ContainerFactory.ts
```diff
@@ -17,13 +20,26 @@ export function createContainer(): Container {
...
     }));
-    const sessionHandler = SessionMiddleware(config, sessionStore);
+    container.bind(SessionStore).toConstantValue(sessionStore);
+    container.bind(SessionMiddleware).toConstantValue(SessionMiddleware(config, sessionStore));
+    container.bind(AuthMiddleware).toConstantValue(new AuthMiddleware());
 
-    const authMiddleware = new AuthMiddleware();
+    const kafkaClient = new kafka.KafkaClient({
+        kafkaHost: getEnvOrDefault('KAFKA_BROKER_ADDR'),
+        requestTimeout: 4000 // 4 seconds
```
n-r-b: should this be configurable? why 4 seconds?

src/controllers/CheckYourAppealController.ts
```diff
@@ -26,7 +34,38 @@ export class CheckYourAppealController extends BaseAsyncHttpController {
...
+            .map(info => info[SignInInfoKeys.UserProfile])
+            .extract() as IUserProfile;
+
+        const appealsData = req.session
+            .chain(_ => _.getExtraData())
+            .chain(data => Maybe.fromNullable(data[AppealKeys.APPEALS_KEY]))
+            .extract() as Appeal;
 
+        try {
+            await this.emailService.send({
```
n-r-b: do we need to check first that the userProfile/appealsData is not `undefined` before sending the email?

src/modules/email-publisher/EmailService.ts
```diff
@@ -0,0 +1,50 @@
...
+            throw new Error('Email must be defined');
+        }
+        const message = this.createMessageFrom(email);
+        return this.sendMessage(message);
+    }
+
+    private async sendMessage(message: Message): Promise<void> {
+        const send = util.promisify(this.producer.send).bind(this.producer)
+        return send([{
+            topic: 'email-send',
```
n-r-b: should we make this topic configurable?

src/ContainerFactory.ts
geek-soft: No clear candidate right now but it seems to be good candidate for reuse.

src/ContainerFactory.ts
geek-soft: 4s seemed sensible to me, original 10s seemed bit too much. I will make it configurable.

src/modules/email-publisher/EmailService.ts
geek-soft: It won't change likely I assume so I decided to keep it hardcoded.

src/controllers/CheckYourAppealController.ts
geek-soft: Double checked with Bart and agreed that in first version any problem will result in an error page being displayed. That will get improved in the future based on the outcome of white-board session.

src/controllers/CheckYourAppealController.ts
geek-soft: You are right, I am waiting for @franciscog-kainos to do it in separate PR. At this point in time I cannot do anything better than it without refactoring unrelated code which I am trying to avoid.

test/ApplicationFactory.ts
geek-soft: Shortened it.

src/views/confirmation.njk
geek-soft: Absolutely valid comment. I thought about it as well and after conversation with Bart we will render error message when something failed. So there will be either email and text highlighting email sent or error message. In the future we might improve it.

src/ContainerFactory.ts
geek-soft: On top of that, the reason I am trying to keep it short is because transition waits for an email to be put into email sending queue. I do not want to wait too long to trigger timeouts on proxy side.

src/controllers/CheckYourAppealController.ts
geek-soft: Valid concern. That you for raising it.

I did not want to do that extra check in the scope of the story. I should really expect data to be there on submission otherwise check your answers page should not even display. I expected that there will be some validation of session state when user visits CheckYourAppealController for the first time. Not sure whether it was missed when we implemented other story or whether we created other story. Anyway we need to address it, ideally as separate ticket or at least PR.

n-r-b: I agree with this

src/ContainerFactory.ts
n-r-b: Ok, makes sense - thanks for making this configurable

PR #90, opened 2020-03-11
feat: submit appeal to mongodb

n-r-b:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1384

### Change description
Added POST to appeals API to store appeal in mongo db permanently 

### Work checklist

### Merge instructions

```
```


src/controllers/CheckYourAppealController.ts
```diff
@@ -46,6 +49,16 @@ export class CheckYourAppealController extends BaseAsyncHttpController {
...
 
+        const accessToken = req.session
+            .chain(_ => _.getValue<ISignInInfo>(SessionKey.SignInInfo))
+            .map(info => info[SignInInfoKeys.AccessToken])
+            .map(token => token?.access_token)
+            .extract() as string;
+
+        const companyNumber: string = appealsData.penaltyIdentifier.companyNumber;
+
+        await this.appealSubmissionService.submitAppeal(appealsData, companyNumber, accessToken);
```
geek-soft: A `appealSubmissionService.submitAppeal(appeals` is looks very repetitive. In Poland we say "Masło maślane" which translated means "Buttery butter" :) There is no need to repeat submit and appeal verbs so many times.

src/controllers/CheckYourAppealController.ts
```diff
@@ -11,11 +11,14 @@ import { BaseAsyncHttpController } from 'app/controllers/BaseAsyncHttpController
...
 import { EmailService } from 'app/modules/email-publisher/EmailService'
+import { AppealSubmissionService } from 'app/service/AppealSubmissionService';
 import { CHECK_YOUR_APPEAL_PAGE_URI, CONFIRMATION_PAGE_URI } from 'app/utils/Paths';
 
 @controller(CHECK_YOUR_APPEAL_PAGE_URI, SessionMiddleware, AuthMiddleware)
 export class CheckYourAppealController extends BaseAsyncHttpController {
-    constructor (@inject(EmailService) private readonly emailService: EmailService) {
+
+    constructor (@inject(EmailService) private readonly emailService: EmailService,
+                 @inject(AppealSubmissionService) private readonly appealSubmissionService: AppealSubmissionService) {
```
geek-soft: PR does not show conflicts but you will have to merge it with my refactoring.

src/service/AppealSubmissionService.ts
```diff
@@ -0,0 +1,37 @@
+import axios, { AxiosRequestConfig } from 'axios';
+import { Appeal } from 'app/models/Appeal';
+
+export class AppealSubmissionService {
+
+    constructor(private readonly appealsApiUrl: string) {
+        this.appealsApiUrl = appealsApiUrl;
+    }
+
+    public async submitAppeal(appealData: Appeal, companyId: string, token: string): Promise<any> {
```
geek-soft: Company number is in `Appeal` object already so I suggest `companyId: string` argument is removed from function declaration. It make me confused and it opens API for misuse allowing API call that passes company A in URI and company B in payload.

src/service/AppealSubmissionService.ts
```diff
@@ -0,0 +1,37 @@
...
+
+        console.log('Making a POST request to ' + uri);
+
+        return await axios
+            .post(uri, appealData, config)
+            .then(response => {
+                console.log(response.data)
+                return response.data
+            })
+            .catch(err => {
```
geek-soft: That adds no value, does it?

src/controllers/CheckYourAppealController.ts
n-r-b: I was trying to think of a good name for this, how about: `appealStorageService.save()`? or any suggestions?

src/controllers/CheckYourAppealController.ts
n-r-b: i have merged your changes in and done some refactoring of my code

src/service/AppealSubmissionService.ts
n-r-b: Good point, make's sense 👍 I was thinking of verifying the company id passed in the URI and the one in the payload is equal on the API side

src/service/AppealSubmissionService.ts
n-r-b: Not right now, I was thinking about handling different error codes but they all return the same generic error page, so this can be removed

akaJSaunders:
I believe there will be some conflicts when I merge the ticket im working on. We have changed the structure of the extra data session a tiny bit, the inclusion of the navigation > permissions field and the change in appeal's level:

export interface ApplicationData {
    appeal: Appeal;
    navigation: {
           permissions: string[];
    }
}

test/controllers/CheckYourAppealController.test.ts
```diff
@@ -1,19 +1,20 @@
...
+import {EmailService} from 'app/modules/email-publisher/EmailService'
+import {AppealStorageService} from 'app/service/AppealStorageService';
+import {CHECK_YOUR_APPEAL_PAGE_URI, CONFIRMATION_PAGE_URI} from 'app/utils/Paths';
 
-import { createApp, getDefaultConfig } from 'test/ApplicationFactory';
-import { createSubstituteOf } from 'test/SubstituteFactory'
-import { createFakeSession } from 'test/utils/session/FakeSessionFactory';
+import {createApp, getDefaultConfig} from 'test/ApplicationFactory';
+import {createSubstituteOf} from 'test/SubstituteFactory'
+import {createFakeSession} from 'test/utils/session/FakeSessionFactory';
```
akaJSaunders: You seem to have removed all the spaces from in between the imports :(

test/controllers/processors/AppealStorageFormSubmissionProcessor.test.ts
```diff
@@ -0,0 +1,70 @@
+import 'reflect-metadata'
+
+import { Arg } from '@fluffy-spoon/substitute';
+import * as assert from 'assert';
+import { Maybe, Session } from 'ch-node-session-handler';
+import { SessionKey } from 'ch-node-session-handler/lib/session/keys/SessionKey';
+import { SignInInfoKeys } from 'ch-node-session-handler/lib/session/keys/SignInInfoKeys';
+import {IAccessToken, ISignInInfo} from 'ch-node-session-handler/lib/session/model/SessionInterfaces';
```
akaJSaunders: space after the {  and before }

test/service/AppealStorageService.test.ts
```diff
@@ -0,0 +1,36 @@
...
+            description: 'they are legit'
+        }
+    }
+};
+
+describe('AppealStorageService', () => {
+
+    describe('saving appeals', () => {
+
+        it('should throw an error when token not defined', () => {
```
akaJSaunders: shouldn't we have a happy path test

test/controllers/CheckYourAppealController.test.ts
n-r-b: must be my .editorconfig 🤔

test/service/AppealStorageService.test.ts
n-r-b: yes we should! i'll add one

src/service/AppealStorageService.ts
```diff
@@ -0,0 +1,35 @@
...
+
+        const uri: string = apiUri + '/companies/' + `${appealData.penaltyIdentifier.companyNumber}`
+            + `/appeals with token: ${token}`;
+
+        console.log('Making a POST request to ' + uri);
+
+        return await axios
+            .post(uri, appealData, config)
+            .then(response => {
+                console.log(response.data);
```
geek-soft: Having just response body printed in logs is not very helpful. Either remove or turn it into useful log that can be used for troubleshooting.

src/controllers/processors/AppealStorageFormSubmissionProcessor.ts
```diff
@@ -0,0 +1,45 @@
...
+            .map(token => token?.access_token as string)
+            .unsafeCoerce();
+
+        const appeal = req.session
+            .chain(_ => _.getExtraData())
+            .map<ApplicationData>(data => data[APPLICATION_DATA_KEY])
+            .map(data => data.appeal as Appeal)
+            .unsafeCoerce();
+
+        console.log(`Saving appeal for userId: ${userId} accessToken: ${accessToken}`);
```
geek-soft: I am against printing access tokens in the logs. It almost like printing user password there. Possession of access token gives a holder access to all APIs.

src/controllers/processors/AppealStorageFormSubmissionProcessor.ts
```diff
@@ -0,0 +1,45 @@
...
+
+        const accessToken = signInInfo
+            .map(info => info[SignInInfoKeys.AccessToken])
+            .map(token => token?.access_token as string)
+            .unsafeCoerce();
+
+        const appeal = req.session
+            .chain(_ => _.getExtraData())
+            .map<ApplicationData>(data => data[APPLICATION_DATA_KEY])
+            .map(data => data.appeal as Appeal)
```
geek-soft: Do you need to do ` as Appeal`? Isn't type known at this stage?

.env.vagrant
```diff
@@ -13,3 +13,5 @@ KAFKA_BROKER_ADDR=chs-kafka:9092
 SC_TEAM_EMAIL='appeals.ch.fake+SC@gmail.com'
 NI_TEAM_EMAIL='appeals.ch.fake+NI@gmail.com'
 DEFAULT_TEAM_EMAIL='appeals.ch.fake+DEFAULT@gmail.com'
+
+APPEALS_API_URL=http://api.chs-dev.internal
```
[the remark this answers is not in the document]
n-r-b: yes it is, confirmed this with Krzys

.env.local
```diff
@@ -13,3 +13,5 @@ KAFKA_BROKER_ADDR=localhost:9092
 SC_TEAM_EMAIL='appeals.ch.fake+SC@gmail.com'
 NI_TEAM_EMAIL='appeals.ch.fake+NI@gmail.com'
 DEFAULT_TEAM_EMAIL='appeals.ch.fake+DEFAULT@gmail.com'
+
+APPEALS_API_URL=http://localhost:9000
```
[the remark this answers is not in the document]
n-r-b: good point - it won't, as a specific header is required for the api... i'll have to have a think about this. The api would just return an exception

src/service/AppealStorageService.ts
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        const config: AxiosRequestConfig = {
+            headers: {
+                Accept: 'application/json',
+                Authorization: 'Bearer ' + token
+            }
+        };
+
+        const apiUri = getEnvOrThrow(`APPEALS_API_URL`);
```
geek-soft: Application should fail on start when config is missing not at runtime. Would it help to move it outside of the class so that config is read as soon as application started?

test/service/AppealStorageService.test.ts
geek-soft: I still cannot see happy path test.

n-r-b: created a test for the happy path

src/service/AppealStorageService.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class AppealStorageService {
+
+    constructor(private readonly uri: string) {
+        this.uri = uri;
+    }
+
+    public async save(appealData: Appeal, token: string): Promise<any> {
+
+        if (token == null) {
```
geek-soft: It would make sense to add null check for appeal arg as well.

And you could rename arg to be just `appeal`.

src/service/AppealStorageService.ts
```diff
@@ -0,0 +1,34 @@
...
+        };
+
+        const uri: string = `${this.uri}/companies/${appealData.penaltyIdentifier.companyNumber}/appeals`;
+
+        console.log('Making a POST request to ' + uri);
+
+        return await axios
+            .post(uri, appealData, config)
+            .then((response: AxiosResponse) => {
+                return response;
```
geek-soft: I am not a big fan of this. Returning whole response unnecessarily leaks HTTP context outside of the service. Service in its purpose should provide encapsulation so that consumers does not have to think in context of HTTP requests. This change makes design rubbish.

Please revert and make service either return resource ID or error which is not HTTP specific.

test/service/AppealStorageService.test.ts
```diff
@@ -0,0 +1,71 @@
...
+                                description: 'they are legit'
+                            }
+                        }
+                    }, {
+                        reqheaders: {
+                            authorization: 'Bearer 123',
+                        },
+                    }
+                )
+                .reply(201, {}, {'location': '/companies/00345567/appeals/555'});
```
geek-soft: Please test other service responses such as 400, 401 and 500.

src/service/AppealStorageService.ts
n-r-b: we are now returning a resource id (location header)

test/service/AppealStorageService.test.ts
n-r-b: Covered all except from a 400 response code as the function validates the token/appeal parameters before passing to the api. API only returns 400 if the data is not sent, but a 422 if data is invalid (which i've covered)

src/service/AppealStorageService.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { Appeal } from 'app/models/Appeal';
+
+export class AppealStorageService {
+
+    constructor(private readonly uri: string) {
+        this.uri = uri;
+    }
+
+    public async save(appeal: Appeal, token: string): Promise<any> {
```
geek-soft: Please change return type to `Promise<string>`

n-r-b: done

PR #104, opened 2020-03-18
feat: add print appeal link

akaJSaunders:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1285

attached bellow is an example of the file being printed:
[GOV.UK - The best place to find government services and information.pdf](https://github.com/companieshouse/lfp-appeals-frontend/files/4350005/GOV.UK.-.The.best.place.to.find.government.services.and.information.pdf)

### Change description

With the JavaScript enabled:
When clicked, the user is presented with a print or download prompt (as implement and set up in user's device).

With JavaScript disabled:
The print link button should be invisible.

### Work checklist

### Merge instructions

```
```


src/controllers/ConfirmationController.ts
```diff
@@ -31,23 +30,13 @@ export class ConfirmationController extends SafeNavigationBaseController<any> {
...
-            .map(penaltyIdentifier => penaltyIdentifier.companyNumber)
-            .extract();
-
-        const userEmail = session
-            .getValue<ISignInInfo>(SessionKey.SignInInfo)
+        const userProfile = session.getValue<ISignInInfo>(SessionKey.SignInInfo)
             .map(info => info[SignInInfoKeys.UserProfile])
-            .map(userProfile => userProfile?.email)
-            .extract();
+            .orDefault({});
```
geek-soft: Behave more expressive there. If user profile is not there blow up. A `unsafeCoerce` will do the job.

src/public/sass/pages/_confirmation-page.scss
```diff
@@ -0,0 +1,18 @@
+.print-summary-body {
+    display: none;
+}
+
+@media print {
```
geek-soft: Can you do something similar to:

```
@include govuk-media-query($media-type: print) {
     .govuk-\!-display-none-print {
       display: none !important;
     }
   }
```

but for screens. Then you just tags elements that should be hidden with one simple class.

src/public/sass/pages/_confirmation-page.scss
```diff
@@ -0,0 +1,18 @@
+.print-summary-body {
```
geek-soft: Remove and use `.govuk-!-display-none` instead. The point is to reuse as much as we can from GDS styles.

src/views/confirmation.njk
```diff
@@ -1,31 +1,54 @@
 {% extends 'layout.njk' %}
 
 {% from 'govuk/components/panel/macro.njk' import govukPanel %}
+{% from 'govuk/components/summary-list/macro.njk' import govukSummaryList %}
 
 {% block beforeContent %}{% endblock %}
 
 {% block content %}
+  <noscript>
```
geek-soft: Would be ideal to place it in `{% block bodyStart %}{% endblock %}` block so it is placed at the beginning of body. It will be called sooner as one benefit and it will mirror what we do in `{% block bodyEnd %}` as second benefit.

src/public/sass/pages/_confirmation-page.scss
```diff
@@ -0,0 +1,14 @@
...
+    .govuk-\!-display-none-screen {
+        display: none !important;
+    }
+}
+
+@include govuk-media-query($media-type: print) {
+    .govuk-footer {
+        display: none !important;
+    }
+    .govuk-panel{
```
geek-soft: space please.

PR #105, opened 2020-03-18
feat: add PIWIK tracking code and define goal on start now button

geek-soft:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1523

### Change description

Adds PIWIK tracking code to application.

In addition to default tracking it also tracks "Start now" button clicks as tracking goal.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/.env.local
+++ b/.env.local
@@ -15,3 +15,5 @@ KAFKA_BROKER_ADDR=localhost:9092
 SC_TEAM_EMAIL='appeals.ch.fake+SC@gmail.com'
 NI_TEAM_EMAIL='appeals.ch.fake+NI@gmail.com'
 DEFAULT_TEAM_EMAIL='appeals.ch.fake+DEFAULT@gmail.com'
+
+CDN_HOST=chs-dev.internal:8080
--- a/src/utils/ConfigLoader.ts
+++ b/src/utils/ConfigLoader.ts
@@ -57,7 +57,7 @@ export const getExpressAppConfig = (directory: string) => (app: express.Applicat
     const url = getEnv('PIWIK_URL');
     const site = getEnv('PIWIK_SITE_ID');
     if (url && site) {
-        app.locals.piwik = { url, site }
+        app.locals.piwik = { url, site };
         app.locals.cdn = {
             host: getEnvOrThrow('CDN_HOST')
         }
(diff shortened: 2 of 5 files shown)


.env.local
```diff
@@ -15,3 +15,5 @@ KAFKA_BROKER_ADDR=localhost:9092
 SC_TEAM_EMAIL='appeals.ch.fake+SC@gmail.com'
 NI_TEAM_EMAIL='appeals.ch.fake+NI@gmail.com'
 DEFAULT_TEAM_EMAIL='appeals.ch.fake+DEFAULT@gmail.com'
+
+CDN_HOST=chs-dev.internal:8080
```
n-r-b: do we need to add this here if it's only used when the PIWIK_URL and PIWIK_SITE_ID are set?

src/utils/ConfigLoader.ts
```diff
@@ -51,6 +52,15 @@ export const getExpressAppConfig = (directory: string) => (app: express.Applicat
                 visuallyHiddenText: accessibleName
             }
         }
+    };
+
+    const url = getEnv('PIWIK_URL');
+    const site = getEnv('PIWIK_SITE_ID');
```
n-r-b: is the PIWIK_SITE_ID the same for all apps?

src/views/landing.njk
```diff
@@ -31,7 +31,8 @@
             text: 'Start now',
             isStartButton: true,
             attributes: {
-              id: 'start-now'
+              id: 'start-now',
```
n-r-b: The acceptance criteria mentions adding this to all continue like buttons, do we need to do this?

geek-soft: AC is wrong in that regard. It looks CH only sets goal on "start now" button

src/utils/ConfigLoader.ts
geek-soft: Nope, every env has different site ID. For example we use 26 in Rebel and 10 in Staging.

.env.local
geek-soft: Good point, not strictly necessary but I know we will have CDN in use very soon.

src/utils/ConfigLoader.ts
n-r-b: ok, i've already seen this has been configured in the global env properties 👍

PR #111, opened 2020-03-25
feature: add upload page

akaJSaunders:
### JIRA link

### Change description

Sub story 1

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "gulp-typescript": "^6.0.0-alpha.1",
     "gulp-watch-sass": "^1.4.1",
     "husky": "^4.2.3",
+    "minimist": "^1.2.5",
     "mocha": "^7.1.1",
     "nock": "^12.0.3",
     "node-sass": "^4.13.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -6200,9 +6200,9 @@
       }
     },
     "minimist": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz",
-      "integrity": "sha1-o1AIsg9BOD7sH7kU9M1d95omQoQ="
+      "version": "1.2.5",
+      "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.5.tgz",
+      "integrity": "sha512-FM9nNUYrRBAELZQT3xeZQ7fmMOBg6nWNmJKTcgsJeaLstP/UODVpGsr5OhXhhXg6f+qtJ8uiZ+PUxkDWcgIXLw=="
     },
     "mixin-deep": {
       "version": "1.3.2",


.env.local
```diff
@@ -10,6 +10,9 @@ DEFAULT_SESSION_EXPIRATION=3600
 
 ACCOUNT_WEB_URL=http://localhost:3001
 
+FILE_TRANSFER_API_HOST='https://49d6uvlsrc.execute-api.eu-west-1.amazonaws.com'
+FILE_TRANSFER_API_KEY='wtG2I1FCuP1GIv3kYNJM53e3bP3YgJuIa2IF0ZDL'
```
geek-soft: Please delete immediately, it should never be added to public repository. Wipe it of the history as well.

.env.vagrant
```diff
@@ -10,6 +10,9 @@ DEFAULT_SESSION_EXPIRATION=3600
 
 ACCOUNT_WEB_URL=http://web.chs-dev.internal:4000
 
+FILE_TRANSFER_API_HOST='https://49d6uvlsrc.execute-api.eu-west-1.amazonaws.com'
+FILE_TRANSFER_API_KEY='wtG2I1FCuP1GIv3kYNJM53e3bP3YgJuIa2IF0ZDL'
```
geek-soft: Same here.

package.json
```diff
@@ -19,14 +19,17 @@
   },
   "dependencies": {
     "@hapi/joi": "^17.1.1",
+    "@types/busboy": "^0.2.3",
```
geek-soft: Types are not dev dependencies. Move to other group.

package.json
```diff
@@ -19,14 +19,17 @@
...
     "avsc": "^5.4.19",
     "axios": "^0.19.2",
     "body-parser": "^1.18.2",
+    "busboy": "^0.3.1",
     "ch-node-session-handler": "git+ssh://git@github.com/companieshouse/node-session-handler.git#c85dad518b56374400ee48ce5340d7fbe24a4a5f",
     "cookie-parser": "^1.4.5",
     "dotenv": "^8.2.0",
     "express": "^4.15.5",
     "express-async-handler": "^1.1.4",
+    "form-data": "^3.0.0",
```
geek-soft: Why do we need that? Cannot we just use JSON literals or https://developer.mozilla.org/en-US/docs/Web/API/FormData?

It feels a bit like overkill.

src/ContainerFactory.ts
```diff
@@ -59,6 +60,9 @@ export function createContainer(): Container {
     container.bind(AppealStorageService).toConstantValue(
         new AppealStorageService(getEnvOrThrow(`APPEALS_API_URL`)));
 
+    container.bind(EvidenceUploadService).toConstantValue(
+        new EvidenceUploadService(getEnvOrThrow(`FILE_TRANSFER_API_HOST`), getEnvOrThrow(`FILE_TRANSFER_API_KEY`)));
```
geek-soft: A `FILE_TRANSFER_API_HOST` is never used by CH. Please reuse `FILE_TRANSFER_API_URL` which contains protocol information.

src/controllers/EvidenceUploadController.ts
```diff
@@ -0,0 +1,38 @@
...
+
+@controller(EVIDENCE_UPLOAD_PAGE_URI, SessionMiddleware, AuthMiddleware)
+export class EvidenceUploadController extends BaseController<OtherReason> {
+    constructor() {
+        super(template, navigation, undefined, undefined,
+            [EvidenceUploadFormSubmissionProcessor]);
+    }
+
+    protected prepareViewModelFromAppeal(appeal: Appeal): Record<string, any> & OtherReason {
+        return appeal.reasons?.other;
```
geek-soft: That seems to make no sense. Evidence view does not need reason details for view rendering.

src/controllers/EvidenceUploadController.ts
```diff
@@ -0,0 +1,38 @@
...
+    previous(): string {
+        return OTHER_REASON_PAGE_URI;
+    },
+    next(): string {
+        return CHECK_YOUR_APPEAL_PAGE_URI;
+    }
+};
+
+@controller(EVIDENCE_UPLOAD_PAGE_URI, SessionMiddleware, AuthMiddleware)
+export class EvidenceUploadController extends BaseController<OtherReason> {
```
geek-soft: Implementation is not quite done at the moment as no continue and continue without files actions are handled.

test/service/EvidenceUploadService.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as FormData from 'form-data';
+import nock = require('nock');
+
+import { EvidenceUploadService } from 'app/service/EvidenceUploadService'
+
+describe('EvidenceUploadService', () => {
+
+    const KEY: string = 'mockKey';
+    const HOST: string = 'http://localhost';
+    const evidenceUploadService = new EvidenceUploadService(HOST , KEY);
```
geek-soft: Service name could be made more generic as in the future it will likely also have file download capability. When renamed it could have been moved to separate PR and merged as a first bit so that Francisco could expand on it by adding download function.

test/service/EvidenceUploadService.test.ts
```diff
@@ -0,0 +1,69 @@
...
+            [undefined, null].forEach(async filename => {
+                try {
+                    await evidenceUploadService.upload(evidence as Buffer, filename!)
+                } catch (err) {
+                    expect(err).to.be.instanceOf(Error)
+                        .and.to.haveOwnProperty('message').equal('File name is missing')
+                }
+            })
+        });
+
```
geek-soft: Unnecessary blank line.

test/service/EvidenceUploadService.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        it('should return 201 status code when evidence uploaded', async () => {
+
+            const EVIDENCE_ID: string = 'Mock ID';
+            const evidence: Buffer = Buffer.from('This is a test');
+
+            const data = new FormData();
```
geek-soft: I do not see any use of it. Please remove.

test/service/EvidenceUploadService.test.ts
```diff
@@ -0,0 +1,69 @@
...
+                    {
+                        reqheaders: {
+                            'x-api-key': KEY
+                        },
+                    }
+                )
+                .reply(201, {id: EVIDENCE_ID});
+
+
+            await evidenceUploadService.upload(evidence, 'test')
```
geek-soft: I prefer

```
const response = await evidenceUploadService.upload(evidence, 'test')
expect(response).to.equal(EVIDENCE_ID);
```

test/service/EvidenceUploadService.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import { EvidenceUploadService } from 'app/service/EvidenceUploadService'
+
+describe('EvidenceUploadService', () => {
+
+    const KEY: string = 'mockKey';
+    const HOST: string = 'http://localhost';
+    const evidenceUploadService = new EvidenceUploadService(HOST , KEY);
+
+    describe('upload file', () => {
```
geek-soft: According to API definition it can also return 415. Looks like we need to handle, only question is when. If we need to handle such cases where MIME was not supported in a nicer way than rendering unexpected error page we should add this to case to the test.

https://github.com/companieshouse/file-transfer-api/blob/master/specs/schema.json#L56

src/controllers/processors/EvidenceUploadFormSubmissionProcessor.ts
```diff
@@ -0,0 +1,77 @@
...
+
+        console.log(req.headers['content-type']);
+
+        const appeal: Appeal = req.session
+            .chain(_ => _.getExtraData())
+            .map<ApplicationData>(data => data[APPLICATION_DATA_KEY])
+            .map(data => data.appeal)
+            .unsafeCoerce();
+
+        console.log(`Uploading file for company number: ${appeal.penaltyIdentifier.companyNumber}` +
```
geek-soft: Let's use CH logger for any logs that will remain in code base for troubleshooting. I am OK for us to use `console.log` temporarily but PR won't be approved until all `console.log` call are eradicated.

test/service/EvidenceUploadService.test.ts
geek-soft: I double checked and it looks like MIME validation is within the scope of the story. It means that 415 must be handled properly and relevant test must be in place before merge.

src/controllers/processors/EvidenceUploadFormSubmissionProcessor.ts
```diff
@@ -0,0 +1,77 @@
...
+            ` and penalty reference: ${appeal.penaltyIdentifier.penaltyReference}`);
+
+        const maxSizeBytes: number = parseInt('4194304', 10);
```
geek-soft: Max file size should be parametrised via ENV variables.

src/controllers/processors/EvidenceUploadFormSubmissionProcessor.ts
```diff
@@ -0,0 +1,77 @@
...
+            });
+
+        busboy.on('finish', () => {
+            console.log('Done parsing form!');
+            // res.writeHead(303, {Connection: 'close', Location: '/'});
+            // res.end();
+        });
+        req.pipe(busboy);
+
+        await id;
```
geek-soft: That seems silly. Processor does not need to return anything as value is not handled anywhere. Processor communicates with outside world through request, response only.

src/controllers/processors/EvidenceUploadFormSubmissionProcessor.ts
```diff
@@ -0,0 +1,77 @@
...
+
+                });
+            });
+
+        busboy.on('finish', () => {
+            console.log('Done parsing form!');
+            // res.writeHead(303, {Connection: 'close', Location: '/'});
+            // res.end();
+        });
+        req.pipe(busboy);
```
geek-soft: Why is `req.pipe(busboy);` done twice?

PR #115, opened 2020-03-30
feat: implement file upload

n-r-b:
### JIRA link

### Change description
Sub story 1

### Work checklist

### Merge instructions

```
```


src/controllers/index.ts
```diff
@@ -1,6 +1,7 @@
 import 'app/controllers/CheckYourAppealController';
 import 'app/controllers/ConfirmationController';
 import 'app/controllers/EntryController'
+import 'app/controllers/EvidenceUploadController';
```
geek-soft, line 4: Even smarter might be not register controllers when feature is off compared to use of middleware :) To be tried in the future.

src/middleware/FileTransferFeatureMiddleware.ts
```diff
@@ -0,0 +1,22 @@
...
+import { loggerInstance } from 'app/middleware/Logger';
+import { getEnv } from 'app/utils/EnvironmentUtils';
+import { ENTRY_PAGE_URI } from 'app/utils/Paths';
+
+@provide(FileTransferFeatureMiddleware)
+export class FileTransferFeatureMiddleware extends BaseMiddleware {
+
+    public handler: RequestHandler = (req: Request, res: Response, next: NextFunction): void => {
+
+        loggerInstance().info(`${FileTransferFeatureMiddleware.name} - handler: request url is ` + req.url);
```
geek-soft, line 14: @n-r-b either you use string interpolation e.g. `${xxx} - ` or string concatenation e.g. `xxx + ' - '` - no point using both.

geek-soft: Another thing, this log is useless.

src/controllers/index.ts
n-r-b: great point! 👍

src/middleware/FileTransferFeatureMiddleware.ts
geek-soft: The reason I am saying useless is that nobody cares whether URI covered by middleware is hit. It is much useful to know that URI who is disabled was triggered.

We need to look at our logging at some point to make it more useful.

PR #116, opened 2020-03-31
feat: add file transfer service

akaJSaunders:
### Work checklist

### Merge instructions

```
```


src/service/FileTransferService.ts
```diff
@@ -0,0 +1,45 @@
...
+import { CREATED } from 'http-status-codes';
+
+export class FileTransferService {
+
+    constructor(private readonly host: string, private readonly key: string) {
+        this.host = host;
+        this.key = key;
+    }
+
+    private uri: string = `${this.host}/dev/files`;
```
franciscog-kainos: Does this work? Or does it have to like:
`
private uri: string;
`
`
constructor(private readonly host: string, private readonly key: string) {
        this.host = host;
        this.key = key;
        this.uri = '${this.host}/dev/files';
    }`

src/service/FileTransferService.ts
```diff
@@ -0,0 +1,46 @@
...
+import { CREATED } from 'http-status-codes';
+
+export class FileTransferService {
+
+    private readonly uri: string;
+
+    constructor(private readonly host: string, private readonly key: string) {
+        this.host = host;
+        this.key = key;
+        this.uri = `${this.host}/dev/files`;
```
geek-soft: Application will be configured with `FILE_TRANSFER_API_URL` which already contains path such as `dev/files`.

test/service/FileTransferService.test.ts
```diff
@@ -0,0 +1,69 @@
...
+                    await evidenceUploadService.upload(evidence as Buffer, filename!)
+                } catch (err) {
+                    expect(err).to.be.instanceOf(Error)
+                        .and.to.haveOwnProperty('message').equal('File name is missing')
+                }
+            })
+        });
+
+
+        it('should return 201 status code when evidence uploaded', async () => {
```
geek-soft: Upload method should handle 415 response code gracefully which is how API communicates that type of the uploaded file is not supported.

Please add test for at least 415 response code as per https://github.com/companieshouse/file-transfer-api/blob/master/specs/schema.json#L56

geek-soft: PS. I am not sure what 404 scenario is, seems bit uncommon to have `the file does not exist` scenario on resource creation.

test/service/FileTransferService.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+            const EVIDENCE_ID: string = 'Mock ID';
+            const evidence: Buffer = Buffer.from('This is a test');
+
+            const data = new FormData();
+            data.append('upload', evidence, {filename: 'test'});
+
+            nock(HOST)
+                .post(URI,
+                    new RegExp(`form-data; name="upload"[^]*${evidence}`,'m'),
```
geek-soft: This is how API call looks like from the HTTP point of view:

```
00c0: Content-Type: multipart/form-data; boundary=--------------------
0100: ----f62dabfe821ce10a
0116:
== Info: Connection state changed (MAX_CONCURRENT_STREAMS == 128)!
=> Send data, 33605 bytes (0x8345)
0000: --------------------------f62dabfe821ce10a
002c: Content-Disposition: form-data; name="upload"; filename="light-9
006c: 4a07e06a1-v2 (1).woff2"
0085: Content-Type: application/octet-stream
```

geek-soft: It might be clearer to do matching based on file name e.g.

```
new RegExp(`form-data; name="upload"; filename="..."`, 'm')
```

test/service/FileTransferService.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import FormData from 'form-data';
+import nock = require('nock');
+
+import { FileTransferService } from 'app/service/FileTransferService'
+
+describe('FileTransferService', () => {
+
+    const KEY: string = 'mockKey';
+    const HOST: string = 'http://localhost';
+    const evidenceUploadService = new FileTransferService(HOST , KEY);
```
n-r-b: please rename this to `fileTransferService`

test/service/FileTransferService.test.ts
akaJSaunders: But I do not know the filename of evidence

PR #121, opened 2020-04-03
feat: validate uploaded evidence file

n-r-b:
### JIRA link

### Change description
Sub story 2

### Work checklist

### Merge instructions

```
```


src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,94 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
         return appeal.reasons?.other;
     }
 
+    private async renderUploadError( appeal: Appeal, text: string ): Promise<void> {
```
franciscog-kainos, line 42: Don't forget to auto format your code
```suggestion
    `private async renderUploadError(appeal: Appeal, text: string): Promise<void> {
```

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,94 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
...
+        const that = this;
+        const validationResult: ValidationResult = new ValidationResult( [
```
franciscog-kainos, line 44:
```suggestion
        const validationResult: ValidationResult = new ValidationResult([
```

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,94 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
...
-                    const id = await that.fileTransferService.upload(request.file.buffer, request.file.originalname);
+                    const maxNumberOfFiles: number = Number(getEnvOrThrow('MAX_NUMBER_OF_FILES'));
 
                     const appeal: Appeal = request.session
                         .chain(_ => _.getExtraData())
                         .map<ApplicationData>(data => data[APPLICATION_DATA_KEY])
                         .map(data => data.appeal)
                         .unsafeCoerce();
 
+                    try{
```
franciscog-kainos, line 71:
```suggestion
                    try {
```

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,94 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
...
+                        await parseFormData(request, response)
+                    } catch (error){
```
franciscog-kainos, line 73:
```suggestion
                    } catch (error) {
```

src/controllers/EvidenceUploadController.ts
franciscog-kainos: And the others...

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,94 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
...
+                        return await that
+                            .renderUploadError(appeal, 'You can only select up to 10 files at the same time');
+                    }
+
+                    let id: string;
+
+                    try {
+                        id = await that.fileTransferService.upload(request.file.buffer, request.file.originalname);
+                    } catch (err) {
+                        if (err.message === 'Request failed with status code 415') {
```
franciscog-kainos, line 98: I can see a pattern in most catch statements.
```
if (err.message === something) {
    doSomething()
    // OR
    returnSomething()
}
```

Could be worth extracting that pattern to avoid repetition

src/service/FileTransferService.ts
```diff
@@ -46,12 +46,6 @@ export class FileTransferService {
                 if (response.status === CREATED && response.data.id) {
                     return response.data.id;
                 }
-            }).catch((err) => {
```
geek-soft, line 49: Why is that removed?

src/utils/MultipartFormDataParser.ts
```diff
@@ -3,10 +3,20 @@ import util from 'util';
...
     multer({
         limits: {
-            fileSize: parseInt(getEnvOrThrow('MAX_FILE_SIZE_BYTES'), 10)
+            fileSize: Number(getEnvOrThrow('MAX_FILE_SIZE_BYTES'))
+        },
+        // @ts-ignore
+        // tslint:disable-next-line: typedef
+        fileFilter (request, file, cb) {
+            if (!supportedFileTypes.includes(file.mimetype)) {
+                return cb(new Error('File not supported'))
```
geek-soft, line 17: Include file's mime type that is not supported in error message.

src/utils/MultipartFormDataParser.ts
```diff
@@ -3,10 +3,20 @@ import util from 'util';
 
 import { getEnvOrThrow } from 'app/utils/EnvironmentUtils';
 
+const supportedFileTypes: string[] = getEnvOrThrow('SUPPORTED_MIME_TYPES').split(',');
+
 export const parseFormData = util.promisify(
     multer({
         limits: {
-            fileSize: parseInt(getEnvOrThrow('MAX_FILE_SIZE_BYTES'), 10)
+            fileSize: Number(getEnvOrThrow('MAX_FILE_SIZE_BYTES'))
```
geek-soft, line 11: What is that change for? Why was that change needed?

src/service/FileTransferService.ts
geek-soft: Please bring that back. We do not want to expose errors from HTTP library outside.

.env.local
```diff
@@ -13,6 +13,8 @@ ACCOUNT_WEB_URL=http://localhost:3001
 FILE_TRANSFER_API_URL=
 FILE_TRANSFER_API_KEY=
 MAX_FILE_SIZE_BYTES=4194304
+SUPPORTED_MIME_TYPES='text/plain,application/msword,application/pdf,image/jpeg,image/png'
```
geek-soft, line 16: What about *.jpg or *.docx?

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,98 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
...
                     appeal.reasons.other.attachments = [...appeal.reasons.other.attachments || [], {
                         id, name: request.file.originalname, contentType: request.file.mimetype, size: request.file.size
                     }];
 
                     await that.persistSession();
 
+                    response.redirect(request.route.path);
+                }
+            },
+            'upload-file-continue': {
```
geek-soft, line 119: I did not expect this. There is built-in capability of BaseController that validates continue buttons, I suggest you use it.

It should be just matter of passing validator instance to base controller.

.env.local
akaJSaunders: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

from my understanding .jpg and jpeg are both captured by  image/jpeg

Screenshot 2020-04-08 at 10 00 05

src/utils/MultipartFormDataParser.ts
akaJSaunders: I was playing around with it and I probably left it like that. But in our case both will return the same, except parseInt takes an extra parameter while Number doesnt

src/utils/MultipartFormDataParser.ts
akaJSaunders: But this error is caught anyway by the controller so that we can display the message: 
"The selected file must be a TXT, DOC, PDF, JPEG or PNG"

src/utils/MultipartFormDataParser.ts
akaJSaunders: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

parseInt apparently can have some unpredictable behaviour, for example If parseInt encounters a character that is not a numeral in the specified radix, it ignores it and all succeeding characters and returns the integer value parsed up to that point. If you want the string to fail with NaN if it has other characters in it, Number() may actually be a better choice.

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,98 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
...
                 async handle(request: Request, response: Response): Promise<void> {
-                    await parseFormData(request, response);
-
-                    if (!request.file) {
-                        response.redirect(request.route.path);
-                        return;
-                    }
 
-                    const id = await that.fileTransferService.upload(request.file.buffer, request.file.originalname);
+                    const maxNumberOfFiles: number = Number(getEnvOrThrow('MAX_NUMBER_OF_FILES'));
```
geek-soft, line 68: Throwing error in method bodies is not desired as it will fail only at the time users use the feature. Configuration should be validated at application start and fail fast. Please move it either outside of controller class or into constructor as last resort. Finally make sure application lunch crashes when value is missing.

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,98 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
...
                     const appeal: Appeal = request.session
                         .chain(_ => _.getExtraData())
                         .map<ApplicationData>(data => data[APPLICATION_DATA_KEY])
                         .map(data => data.appeal)
                         .unsafeCoerce();
 
+                    try {
+                        await parseFormData(request, response)
+                    } catch (error) {
+                        if (error.message === 'File not supported') {
```
geek-soft, line 79: I am bit sceptical for implementing application logic against error messages. Library uses codes in Multer error messages and I feel it is bit safer to use these codes as codes are supported to be for machines while messages are supported to be for people.

See: https://github.com/expressjs/multer/blob/805170c61530e1f1cafd818c9b63d16a9dd46c36/lib/multer-error.js#L3

I suggest you build your condition around error codes.

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,98 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
...
+                            return await that
+                                .renderUploadError(appeal, fileNotSupportedError);
+                        } else if (error.message === 'File too large') {
```
geek-soft, line 82: Switch expression would look better here .e.g.

```
switch (error.code) {
  case 'LIMIT_FILE_SIZE': await that.renderUploadError(appeal, fileTooLargeError); break;
  case 'FILTER_FILE_TYPE': await that.renderUploadError(appeal, fileNotSupportedError); break;
}
```

where `FILTER_FILE_TYPE` is custom code throw by file type filter.

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,98 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
         return appeal.reasons?.other;
     }
 
+    private async renderUploadError(appeal: Appeal, text: string): Promise<void> {
+        const that = this;
```
geek-soft, line 43: I doubt `that = this` technique is needed here. Cause it is private method on the controller `this` should have correct value.

src/utils/MultipartFormDataParser.ts
```diff
@@ -3,10 +3,20 @@ import util from 'util';
 
 import { getEnvOrThrow } from 'app/utils/EnvironmentUtils';
 
+const supportedFileTypes: string[] = getEnvOrThrow('SUPPORTED_MIME_TYPES').split(',');
```
geek-soft, line 6: Please double check that application does not start without `SUPPORTED_MIME_TYPES` env var.

src/utils/MultipartFormDataParser.ts
geek-soft: For consistency and in line with my other comment advising to use message codes I suggest we throw Multer exception here with custom code and custom message.

src/controllers/EvidenceUploadController.ts
```diff
@@ -34,32 +39,98 @@ export class EvidenceUploadController extends BaseController<OtherReason> {
...
+                            return await that
+                                .renderUploadError(appeal, fileNotSupportedError);
+                        } else if (error.message === 'File too large') {
+                            return await that
+                                .renderUploadError(appeal, fileTooLargeError);
+                        }
+                    }
+
+                    if (!request.file) {
+                        response.redirect(MOVED_TEMPORARILY, request.route.path);
```
geek-soft, line 89: Shouldn't we render validation error now when we have validation? It is quite surprising for page to reload. Please get in touch with Bart or UX about this. Prototype does not have validation message cause it uses JS version where upload is triggered on file selection automatically but here we should have some validation not a reload IMO.

test/controllers/PenaltyDetailsController.test.ts
```diff
@@ -21,7 +21,7 @@ const config = getDefaultConfig();
 
 describe('PenaltyDetailsController', () => {
 
-    const navigation = {} as Navigation;
+    const navigation: Navigation = {} as Navigation;
```
geek-soft, line 24: What do you think about making these kind of changes in separate PR so they do not distract from the purpose of PR / user story?

src/controllers/EvidenceUploadController.ts
akaJSaunders: that makes it so much better

test/controllers/CheckYourAppealController.test.ts
```diff
@@ -20,7 +20,7 @@ import { createFakeSession } from 'test/utils/session/FakeSessionFactory';
 
 const config = getDefaultConfig();
 
-const appeal = {
```
geek-soft, line 23: Same here, may I ask you to move that to separate PR. It feels like distraction :(

src/views/evidence-upload.njk
```diff
@@ -22,14 +30,14 @@
           {% else %}
             Add a document
           {% endif %}
-        {% endset %}
+        {% endset -%}
```
geek-soft, line 33: What is that change doing?

src/utils/MultipartFormDataParser.ts
geek-soft: I understand it error is caught but good practice says that error messages should be meaningful and helpful in troubleshooting. You cannot expect that every single error will be handled, some errors might slip and then good error message is essential in understanding what happened.

src/controllers/EvidenceUploadController.ts
akaJSaunders: Will do

.env.local
geek-soft: In that case please double check that `*.jpg` files can be handled through `image/jpeg` mime type.

src/controllers/EvidenceUploadController.ts
akaJSaunders: I put it outside the controller and it crashes

.env.local
akaJSaunders: The screenshot above demonstrates it

test/controllers/PenaltyDetailsController.test.ts
n-r-b: oops, sorry this was me! i took note of the suggestions you made last week :-) i will remove these for now and we can add them back in the future

src/service/FileTransferService.ts
n-r-b: i've put this code back. We are also catching these errors in the controller so that we can render a suitable error on the page for unsupported mime type

src/controllers/EvidenceUploadController.ts
n-r-b: ok, will give this a go

n-r-b: I've created a validator instance for the button, was very simple to do :-)

src/controllers/EvidenceUploadController.ts
```diff
@@ -24,36 +32,100 @@ const navigation = {
...
     }
 
+    private async renderUploadError(appeal: Appeal, text: string): Promise<void> {
+        const validationResult: ValidationResult = new ValidationResult([
+            new ValidationError('file', text)
+        ]);
+
+        return await this.renderWithStatus(UNPROCESSABLE_ENTITY)(
+            this.template, {
+                ...this.prepareViewModelFromAppeal(appeal),
```
geek-soft, line 72: Is "continue without docs" link working after validation error is triggered?

src/controllers/EvidenceUploadController.ts
```diff
@@ -24,36 +32,100 @@ const navigation = {
...
+                ...this.httpContext.request.body,
+                validationResult
+            }
+        );
+    }
+
     protected getExtraActionHandlers(): Record<string, FormActionHandler | FormActionHandlerConstructor> {
         const that = this;
+
+        const fileTooLargeError: string = 'File size must be smaller than 4MB';
```
geek-soft, line 82: I see bit of maintainability issue where file limit env var and error message might get out of sync.

To not add more complexity I suggest we track it tech debt in form of task. Bart should have created tech debt epic for us.

src/controllers/EvidenceUploadController.ts
```diff
@@ -24,36 +32,100 @@ const navigation = {
...
+        const fileNotSupportedError: string = 'The selected file must be a TXT, DOC, PDF, JPEG or PNG';
+        const tooManyFilesError: string = 'You can only select up to 10 files at the same time';
```
geek-soft, line 84: I guess error message might reuse `MAX_NUMBER_OF_FILES ` env variable as this is simple interpolation here.

geek-soft: Or just add it to tech debt as the other thing.

src/controllers/EvidenceUploadController.ts
```diff
@@ -24,36 +32,100 @@ const navigation = {
...
+                    let id: string;
+
+                    try {
+                        id = await that.fileTransferService.upload(request.file.buffer, request.file.originalname);
+                    } catch (err) {
+                        if (err.message === 'Unsupported file type') {
+                            return await that
+                                .renderUploadError(appeal, fileNotSupportedError);
+                        } else {
+                            throw new Error(err.message)
```
geek-soft, line 125: I guess you can just re-throw error e.g. `throw err`. I see no value in wrapping it in another error.

src/controllers/EvidenceUploadController.ts
akaJSaunders: Good spot, I was also wondering about it at the time, but got distracted

src/controllers/EvidenceUploadController.ts
akaJSaunders: Yes

PR #122, opened 2020-04-03
Lfa 1446/download file

franciscog-kainos:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1466

### Change description
Implemented auto-download feature from prompt screen and download link.

Please approve https://github.com/companieshouse/cdn.ch.gov.uk/pull/389 first which is the JS file that will be sent from the CDN to the client for auto download to work.

### Work checklist

### Merge instructions

```
```


src/utils/Paths.ts
```diff
@@ -15,3 +15,4 @@ export const OTHER_REASON_PAGE_URI = `${CATEGORY_PREFIXES.OTHER_REASON}/reason-o
 export const EVIDENCE_UPLOAD_PAGE_URI = `${ROOT_URI}/evidence-upload`;
 export const CHECK_YOUR_APPEAL_PAGE_URI = `${ROOT_URI}/check-your-answers`;
 export const CONFIRMATION_PAGE_URI = `${ROOT_URI}/confirmation`;
+export const DOWNLOAD_FILE_URI = `${ROOT_URI}/download`
```
n-r-b, line 18: can you rename this to DOWNLOAD_FILE_PAGE_URI to be consistent with the others?

src/controllers/FileDownloadController.ts
```diff
@@ -0,0 +1,38 @@
...
+
+@controller(DOWNLOAD_FILE_URI)
+export class FileDownloadController extends BaseAsyncHttpController {
+
+    constructor(@inject(FileTransferService) private readonly fileTransferService: FileTransferService) {
+        super();
+    }
+
+    @httpGet('/prompt/:fileId')
+    public renderPrompt(@response() res: Response, @requestParam('fileId') fileId: string): void {
```
n-r-b, line 21: is the response decorator needed?

src/controllers/FileDownloadController.ts
```diff
@@ -0,0 +1,38 @@
...
+        res.render(template, { fileId });
+    }
+
+    @httpGet('/data/:fileId/download')
+    public async download(@response() res: Response, @requestParam('fileId') fileId: string): Promise<void> {
```
n-r-b, line 26: is the response decorator needed?

src/controllers/FileDownloadController.ts
```diff
@@ -0,0 +1,38 @@
...
+
+        try {
+            await this.fileTransferService.download(fileId, res);
+            res.status(OK);
+        } catch (err) {
+            loggerInstance().error(`${err} - at - ${FileDownloadController.name}.download`);
```
n-r-b, line 32: do we need to catch the error? can we not default to the ErrorHandler.ts?

src/models/FileMetada.ts
```diff
@@ -0,0 +1,11 @@
+export type FileMetada = {
```
n-r-b, line 1: was this supposed to be `FileMetadata`? :D

src/service/FileTransferService.ts
```diff
@@ -42,10 +47,62 @@ export class FileTransferService {
...
-                    throw new Error('Unsupported file type')
+                    throw new Error('Unsupported file type');
                 } else {
-                    throw new Error(err.message)
+                    throw new Error(err.message);
                 }
             });
     }
+
+    async fileMetada(fileId: string): Promise<FileMetada> {
```
n-r-b, line 57: `fileMetadata` here too?

geek-soft, line 57: I suggest we keep using verbs in function names.

n-r-b, line 57: What happens if the fileId is null?

src/service/FileTransferService.ts
```diff
@@ -42,10 +47,62 @@ export class FileTransferService {
...
+        const config: AxiosRequestConfig = {
+            headers: {
+                'x-api-key': this.key
+            },
+        };
+
+        return await axios
+            .get<FileMetada>(`${this.url}/${fileId}`, config)
+            .then(_ => _.data)
```
geek-soft, line 66: May I suggest we stop using `_` as names? There was time I liked that approach especially working with other programming languages but at the end of the day names matters and `_` is against good naming conventions.

src/controllers/FileDownloadController.ts
franciscog-kainos: yes I'm calling `res.render`

src/controllers/FileDownloadController.ts
franciscog-kainos: What do you actually mean by, is it needed?

src/service/FileTransferService.ts
```diff
@@ -42,10 +47,62 @@ export class FileTransferService {
...
+            headers: {
+                'x-api-key': this.key
+            },
+            responseType: 'stream'
+        };
+
+        return axios.get<stream.Readable>(url, config)
+            .then(async (axiosResponse: AxiosResponse<stream.Readable>) => {
+
+                if ((res as Response).setHeader) {
```
geek-soft, line 85: When I saw method signature taking writable stream I thought it is OK but service should know be independent from express.js and as such should know nothing about response.

Please either return readable stream from `download` function or set response headers outside of the download method. This service conceptually close to client library. Setting response headers should be on the controller side.

src/controllers/FileDownloadController.ts
geek-soft: I though we could just use `this.httpContext` for consistency but it is not massive issue although I would not like to start using `@response()` in more places than just here. We settled on using `httpContext`.

src/service/FileTransferService.ts
franciscog-kainos: That sounds good to me

src/controllers/FileDownloadController.ts
geek-soft: The same here I though about using `this.httpContext.response`.

src/service/FileTransferService.ts
franciscog-kainos: You can't pass a null to a string arg

franciscog-kainos: oopsy

src/controllers/FileDownloadController.ts
franciscog-kainos: We need to catch the error to render the error page with some added useful message rather than the usual 404

src/models/FileMetada.ts
```diff
@@ -0,0 +1,11 @@
...
+    av_status: 'scanned' | 'not-scanned';
+    content_type: string;
+    id: string;
+    links: {
```
geek-soft, line 5: Shall we get rid of this? We take control of how resource is being download so it becomes needed.

src/controllers/FileDownloadController.ts
franciscog-kainos: I just used this because I'm not using the BaseController

src/controllers/FileDownloadController.ts
franciscog-kainos: I can change it for consistency then

src/views/download-file.njk
```diff
@@ -0,0 +1,32 @@
+{% extends 'layout.njk' %}
+
+{% block pageTitle %}
+  {% if errorList.length > 0 %}
+    Error:
+  {% endif %}
+  Sign in
+{% endblock %}
+{% block beforeContent %}
```
geek-soft, line 9: That seems unneeded, layout template comes with phase banner configured with right feedback link.

src/views/download-file.njk
```diff
@@ -0,0 +1,32 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl">Your document download will start soon</h1>
+      <p class="govuk-body-l">
+        <a href="{{ paths.DOWNLOAD_FILE_URI }}/data/{{ fileId }}/download" id="downloadLink" class="govuk-link">Click here to download your document</a>
+        if the download does not start in 10 seconds.</p>
+    </div>
+  </div>
+{% endblock %}
+{% block bodyEnd %}
```
geek-soft, line 30: Will it override `bodyEnd` block in parent template containing GOV.UK and PIWIK scripts?

src/controllers/FileDownloadController.ts
```diff
@@ -0,0 +1,38 @@
...
+            return await this.renderWithStatus(err.status)('error', { message: err.message });
```
geek-soft, line 33: Generic error handling should handle these errors. Generally there should be 1-to-1 relationship between controller and template.

src/models/FileMetada.ts
franciscog-kainos: remove links?

geek-soft: Yeh, we will never use it.

src/service/FileTransferService.ts
geek-soft: I am bit on the cross road here. On one hand TS won't allow passing null or undefined if strict type checking is enabled. On the other hand source is transpired to JS which does not care about types and for that reason I tend to have null checks to fail fast

geek-soft: Maybe it would be right to have following guidance:
- if code is internal (will never be exposed outside) we relay on TS type checking
- if code is candidate for extraction to NPM like email service then we add extra safety net for JS only projects?

src/controllers/FileDownloadController.ts
geek-soft: `BaseAsyncHttpController` extends `BaseHttpController` which has `httpContext`

src/controllers/FileDownloadController.ts
geek-soft: Generic error handler can be more flexible in taking heading from error if present

src/service/FileTransferService.ts
geek-soft: I am bit worried that it is so simple to disable string type checks :(

src/controllers/index.ts
```diff
@@ -2,8 +2,10 @@ import 'app/controllers/CheckYourAppealController';
 import 'app/controllers/ConfirmationController';
 import 'app/controllers/EntryController'
 import 'app/controllers/EvidenceUploadController';
```
geek-soft, line 4: At some point we might look in to consistent names. Either we decide to use evidence or file. Does not need to be this PR though.

src/service/FileTransferService.ts
```diff
@@ -42,10 +47,62 @@ export class FileTransferService {
...
+            headers: {
+                'x-api-key': this.key
+            },
+        };
+
+        return await axios
+            .get<FileMetada>(`${this.url}/${fileId}`, config)
+            .then(_ => _.data)
+            .catch(_ => {
+                throw new FileNotFoundError(fileId, _.message);
```
geek-soft, line 68: An API should also handle other scenarios e.g. non 5xx response. We would not like service classify 5xx responses as FileNotFoundError. I guess we can have check whether 404 for returned and throw FileNotFoundError and generic error in other cases (fallback).

src/service/FileTransferService.ts
franciscog-kainos: Just like in Java when you compile to src code, you can use a bytecode editor to do silly things like pass null ref to string args. It depends on how defensively we want to code. I think that guidance is fair.

src/service/FileTransferService.ts
```diff
@@ -42,10 +47,62 @@ export class FileTransferService {
...
+                    response.setHeader('Content-Type', axiosResponse.headers['content-type']);
+                    response.setHeader('Content-Length', axiosResponse.headers['content-length']);
+                    response.setHeader('Content-Disposition', axiosResponse.headers['content-disposition']);
+
+                }
+                await this.pipeDataIntoStream(axiosResponse, res);
+
+            })
+            .catch(_ => {
+                throw new FileDownloadError(fileId, _.message);
```
geek-soft, line 96: Same here, 5xx group of responses should be handled differently.

src/service/error/AbstractFileError.ts
```diff
@@ -0,0 +1,10 @@
+export abstract class AbstractFileError extends Error {
+    constructor(className: string,
```
geek-soft, line 2: I do not think we need class name. In error handler we can use `err.constructor.name` where `err` is an instance of error. It should give us same result as `FileDownloadError.name` I suspect.

src/service/error/AbstractFileError.ts
```diff
@@ -0,0 +1,10 @@
...
+                fileId: string,
+                messageProd: (fileId: string) => string,
```
geek-soft, line 4: May I ask you to use full name to avoid confusion. I suspect you wanted this to be message supplier or producer.

src/service/error/AbstractFileError.ts
franciscog-kainos: I'll have a look at this

src/views/download-file.njk
franciscog-kainos: Now that I added {{ super() }} it won't. Thanks for the spot

src/service/FileTransferService.ts
```diff
@@ -42,10 +46,75 @@ export class FileTransferService {
...
+            .on('finish', resolve)
+            .on('error', reject));
+    }
+
+    private getErrorFrom(err: any, fileId: string): Error {
+
+        if (err.isAxiosError) {
+            switch (err.response.status) {
+                case NOT_FOUND:
+                    return new FileNotFoundError(fileId, err.message);
```
geek-soft, line 110: Can you give me example of `err.message` value in this scenario?  Error log `File ${f} not found.` already explains what happened so I wonder how much more useful it gets.

src/views/download-file.njk
```diff
@@ -0,0 +1,22 @@
+{% extends 'layout.njk' %}
+
+{% block pageTitle %}
+  {% if errorList.length > 0 %}
```
geek-soft, line 4: What is that for? I cannot see any `errorList` being set.

src/views/download-file.njk
```diff
@@ -0,0 +1,22 @@
...
+    Error:
+  {% endif %}
+  Sign in
```
geek-soft, line 7: This is quite surprising especially that user will be signed in already at this point. Is that what prototype says? If so get in touch with UX to raise question whether it is correct.

src/controllers/EvidenceDownloadController.ts
```diff
@@ -0,0 +1,45 @@
...
+const template = 'download-file';
+
+@controller(DOWNLOAD_FILE_PAGE_URI)
+export class EvidenceDownloadController extends BaseAsyncHttpController {
+
+    constructor(@inject(FileTransferService) private readonly fileTransferService: FileTransferService) {
+        super();
+    }
+
+    @httpGet('/prompt/:fileId')
```
geek-soft, line 20: Do we have that prompt screen in prototype? If so what is the URL?

src/controllers/EvidenceDownloadController.ts
```diff
@@ -0,0 +1,45 @@
...
+import { inject } from 'inversify';
+import { controller, httpGet, requestParam } from 'inversify-express-utils';
+import { BaseAsyncHttpController } from './BaseAsyncHttpController';
+
+import { FileTransferService } from 'app/service/FileTransferService';
+import { DOWNLOAD_FILE_PAGE_URI } from 'app/utils/Paths';
+
+const template = 'download-file';
+
+@controller(DOWNLOAD_FILE_PAGE_URI)
```
geek-soft, line 13: Please hide it behind feature flag, same as other controller.

src/service/FileTransferService.ts
franciscog-kainos: `err.message` contains the actual axios err message which may be useful to have in the object for debugging purposes. The `File ${f} not found.` will be displayed on the client rather than a complicated and unnecessarily detailed message

franciscog-kainos: This also explains the purpose of `extraData`

geek-soft: Can you please give me an example for message from Axios? I'd love to see how it does look like.

test/service/FileTransferService.test.ts
```diff
@@ -74,9 +92,84 @@ describe('FileTransferService', () => {
...
+            } catch (err) {
+                expect(err.constructor.name).to.eq(FileNotFoundError.name);
+            }
+
+        });
+        it('should get file metada if it exists', async () => {
+
+            createGetNockRequest(fileMetadaUrl).reply(200, expectedMetada);
+
+            await fileTransferService.getFileMetadata(fileID).then(_ => assert.deepEqual(_, expectedMetada));
```
geek-soft, line 120: For consistency please extract response to variable `const fileMetadata = ...` and compare it on the next line. Please also use `chai.expect` instead of assertions` as this is our main assertion library.

test/service/FileTransferService.test.ts
```diff
@@ -74,9 +92,84 @@ describe('FileTransferService', () => {
...
+            createGetNockRequest(fileMetadaUrl).reply(404);
+
+            try {
+                await fileTransferService.getFileMetadata(fileID);
+            } catch (err) {
+                expect(err.constructor.name).to.eq(FileNotFoundError.name);
+            }
+
+        });
+        it('should get file metada if it exists', async () => {
```
geek-soft, line 116: Please break line for readability.

test/service/FileTransferService.test.ts
```diff
@@ -74,9 +92,84 @@ describe('FileTransferService', () => {
...
+    describe('Download a file', async () => {
+        const readFile = promisify(Fs.readFile);
+
+        const fileToStream = 'package.json';
+        const fileToStreamPath = `./${fileToStream}`;
+        const downloadFileName = 'hello.txt';
+        const downloadedDirPath = `./test`;
+        const downloadedFilePath = `${downloadedDirPath}/${downloadFileName}`;
+        const downloadUrl = `${URI}/${fileID}/download`;
+        const fileDataBuffer = Fs.createReadStream(fileToStream);
```
geek-soft, line 133: All that reading from file is bit over complicated. I suggest simple `Readable.from(string)`. is similar to how upload test operates on buffer.

src/controllers/EvidenceDownloadController.ts
franciscog-kainos: `host:port/download`

src/service/FileTransferService.ts
franciscog-kainos: image

src/controllers/EvidenceDownloadController.ts
```diff
@@ -0,0 +1,44 @@
...
+        super();
+    }
+
+    @httpGet('/prompt/:fileId')
+    public renderPrompt(@requestParam('fileId') fileId: string): void {
+        this.httpContext.response.render(template, { fileId });
+    }
+
+    @httpGet('/data/:fileId/download')
+    public async download(@requestParam('fileId') fileId: string): Promise<any> {
```
geek-soft, line 24: Please change return type to `Promise<void>`.

src/controllers/EvidenceDownloadController.ts
```diff
@@ -0,0 +1,44 @@
...
+        const metadata = await this.fileTransferService.getFileMetadata(fileId);
+        res.setHeader('content-disposition', `attachment; filename=${metadata.name}`);
+
+        const readable = await this.fileTransferService.download(fileId);
+
+        return await this.pipeDataIntoStream(readable, res);
+
+    }
+
+    private async pipeDataIntoStream(readable: Readable, writable: Writable): Promise<void> {
```
geek-soft, line 37: Does piping have to be promisified? Wouldn't it be enough to call `readable.pipe(writable)` directly in download method?

src/controllers/EvidenceDownloadController.ts
geek-soft: Thank you

src/controllers/EvidenceDownloadController.ts
franciscog-kainos: image

Short answer, yes.

test/controllers/EvidenceDownloadController.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const metadata: FileMetadata = {
+        av_status: 'scanned',
+        content_type: contentType,
+        id: FILE_ID,
+        name: 'hello.txt',
+        size: contentLength
+    };
+
+    const fileTransferServiceProxy = (downloadResult: Promise<Readable>): FileTransferService => {
```
geek-soft, line 37: Out of curiosity, why didn't you use SubstituteJs?

src/controllers/EvidenceDownloadController.ts
geek-soft: Alright, I can swallow that sour pill and we can improve it later ;)

test/controllers/EvidenceDownloadController.test.ts
franciscog-kainos: At one point in time I had to actually use Proxy to do something custom like adding a custom callback when I had one. But now, yes I could just use SubstituteJs

geek-soft: I suggest we use SubstituteJS fo consistency to not set an exception that could be copied over and over and spread like virus :)

PR #123, opened 2020-04-03
feat: add ability to remove uploaded documents

geek-soft:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1279

### Work checklist

### Merge instructions

```
```


src/models/chunks/YesNo.schema.ts
```diff
@@ -0,0 +1,18 @@
...
+        .messages({
+            'any.required': errorMessage,
+            'any.only': errorMessage,
+            'string.base': errorMessage,
+            'string.empty': errorMessage
+        })
+        .options({
+            abortEarly: true
+        })
+};
```
akaJSaunders, line 18: Why did you call it Chunks? Doesn't mean much to me, could it be renamed?

src/service/FileTransferService.ts
```diff
@@ -1,14 +1,18 @@
...
+import { CREATED, NOT_FOUND, UNSUPPORTED_MEDIA_TYPE } from 'http-status-codes';
 
 import { loggerInstance } from 'app/middleware/Logger';
 
 export class FileTransferService {
 
     constructor(private readonly url: string, private readonly key: string) {
-        this.url = url;
-        this.key = key;
+        if (url == null) {
```
akaJSaunders, line 10: I'v noticed that you have been using '==' rather than '===' throughout your implementation, what is the reason behind it?  From my understanding the triple equals compared both the type and the value, and is therefore more strict. Shouldn't we by default use it?

src/service/FileTransferService.ts
```diff
@@ -48,4 +52,33 @@ export class FileTransferService {
...
+                }
+                throw new Error(`File ${fileId} cannot be deleted due to error: ${(err.message || 'unknown error').toLowerCase()}`)
+            });
+    }
+
+    private prepareHeaders(): Record<string, string> {
+        return {
+            'x-api-key': this.key
+        }
+    }
```
akaJSaunders, line 83: Considering that the request only takes one header and it is unlikely that the API will be changed, why do we need this extra method?

test/service/FileTransferService.test.ts
```diff
@@ -79,4 +80,67 @@ describe('FileTransferService', () => {
...
+                .delete(`${URI}/${fileId}`)
+                .reply(404, {
+                    message: 'file not found'
+                });
+
+            try {
+                await fileTransferService.delete(fileId);
+                assert.fail('Test should failed while it did not')
+            } catch (err) {
+                expect(err.message).to.contain(`File ${fileId} cannot be deleted because it does not exist`);
```
n-r-b, line 111: can we 'expect' the error code here too?

src/controllers/EvidenceRemovalController.ts
```diff
@@ -0,0 +1,86 @@
...
+@controller(EVIDENCE_REMOVAL_PAGE_URI, SessionMiddleware, AuthMiddleware, FileTransferFeatureMiddleware)
+export class EvidenceRemovalController extends BaseController<any> {
+    constructor() {
+        super(template, navigation, new FormValidator(schema), undefined, [Processor]);
+    }
+
+    protected prepareViewModelFromAppeal(appeal: Appeal): any {
+        const attachment: Attachment = findAttachment(appeal, this.httpContext.request.query.f);
+
+        return { fileId: attachment.id, fileName: attachment.name };
```
n-r-b, line 84: could we have just returned the Attachment object here?

src/models/chunks/YesNo.ts
```diff
@@ -0,0 +1,4 @@
+export enum YesNo {
```
n-r-b, line 1: Do we really need an enum to hold yes/no values?

src/controllers/EvidenceRemovalController.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        const appeal: Appeal = request.session
+            .chain(session => session.getExtraData())
+            .map<ApplicationData>(data => data[APPLICATION_DATA_KEY])
+            .map(applicationData => applicationData.appeal)
+            .unsafeCoerce();
+
+        const attachment = findAttachment(appeal, request.body.fileId);
+        await this.fileTransferService.delete(attachment.id);
+        appeal.reasons.other.attachments!.splice(appeal.reasons.other.attachments!.indexOf(attachment), 1)
```
n-r-b, line 70: would a debug statement be helpful here as we are updating the session?

src/models/chunks/YesNo.schema.ts
geek-soft: Cause I could not find better name. The intention is to have field level reusable schemas for things like Yes/No, penalty reference, company number and so on. How would you name it?

src/service/FileTransferService.ts
geek-soft: To cut on repetition.

test/service/FileTransferService.test.ts
geek-soft: Error throw by an API does not come with error code. It hides response code concern from service consumers.

src/models/chunks/YesNo.ts
geek-soft: Do we really need...? No, it can be implemented without enum. My reason for using enum is to reduce chance of making mistakes in situation where only two values 'yes' and 'no' are correct.

What is your concern for having an enum?

src/controllers/EvidenceRemovalController.ts
geek-soft: I am not big fan of debug statements to be honest. In my opinion there is no need to have debug statements in the code base that is properly tested. I'd expect to have only debug statements in certain places for example in session handler logging what is being saved in Redis.

src/controllers/EvidenceRemovalController.ts
geek-soft: It is possible. Dedicated view model played role of anti-corrosion layer but we do not use that technic in other places so I am happy to simplify if for now.

src/controllers/EvidenceRemovalController.ts
```diff
@@ -0,0 +1,86 @@
...
+            return;
+        }
+
+        const appeal: Appeal = request.session
+            .chain(session => session.getExtraData())
+            .map<ApplicationData>(data => data[APPLICATION_DATA_KEY])
+            .map(applicationData => applicationData.appeal)
+            .unsafeCoerce();
+
+        const attachment = findAttachment(appeal, request.body.fileId);
```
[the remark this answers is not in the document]
geek-soft: Added. Thanks.

src/service/FileTransferService.ts
```diff
@@ -48,4 +52,33 @@ export class FileTransferService {
...
+        if (fileId == null) {
+            throw new Error('File ID is missing');
+        }
+
+        const config: AxiosRequestConfig = {
+            headers: {
+                ...this.prepareHeaders()
+            }
+        };
+
```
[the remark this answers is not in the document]
geek-soft: I believe that better solution is to throw meaningful error. What would you use `Making a POST request to ${this.url}` log for? Especially that `${this.url}` evaluates to known config value?

src/models/chunks/YesNo.schema.ts
geek-soft: What about `core`?

src/service/FileTransferService.ts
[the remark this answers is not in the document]
geek-soft: I will rewrite logging and error handling in separate PR following this one.

src/models/chunks/YesNo.schema.ts
geek-soft: Renamed to `fields` as per our conversation.

src/service/FileTransferService.ts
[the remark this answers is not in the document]
n-r-b: it was only for reference, to show us that the call to the service had taken place and for a specific file

src/models/chunks/YesNo.ts
n-r-b: I was just curious as the values are simply 'yes' or 'no'. Happy to stick with the enum though 👍

PR #129, opened 2020-04-08
refactor: refactor file transfer service

geek-soft:
### JIRA link

None

### Change description

This PR does:
- move service into modules directory
- implements consistent error handling
- add support for scenarios where file is either infected or not scanned
- implements consistent unit tests

### Work checklist

### Merge instructions

```
```


src/modules/file-transfer-service/FileTransferService.ts
```diff
@@ -120,15 +101,22 @@ export class FileTransferService {
...
     }
 
-    private getErrorFrom(err: any, fileId: string): Error {
-
-        if (err.isAxiosError) {
-            switch (err.response.status) {
-                case NOT_FOUND:
-                    return new Error(`File ${fileId} not found.`);
+    private handleResponseError(operation: 'upload' | 'metadata retrieval' | 'download' | 'deletion', subject: string):
+        (err: AxiosError) => never {
```
n-r-b, line 105: does this mean that the following (within `{ }`) should 'never' occur?

src/modules/file-transfer-service/FileTransferService.ts
```diff
@@ -120,15 +101,22 @@ export class FileTransferService {
...
+        // tslint:disable: max-line-length
+        return (err: AxiosError) => {
+            if (err.isAxiosError) {
```
n-r-b, line 108: wouldn't the `err` always be of type `AxiosError`?

src/modules/file-transfer-service/FileTransferService.ts
geek-soft: Never type is required by `catch` function. I started with `void` but `never` was required.

src/modules/file-transfer-service/FileTransferService.ts
geek-soft: Hmm, not sure, Francisco started this convention and I think it gives us extra safety.

src/modules/file-transfer-service/FileTransferService.ts
```diff
@@ -120,15 +101,22 @@ export class FileTransferService {
...
+                switch (err.response!.status) {
+                    case FORBIDDEN:
```
n-r-b, line 110: To me, 'FORBIDDEN' means unauthorised, not 'in progress'....however i don't really know of a more suitable error code for this

src/modules/file-transfer-service/FileTransferService.ts
geek-soft: This is what docs say: https://github.com/axios/axios#handling-errors

It does not explain what errors you can get but with HTTP which is complex topic I feel more safe with that condition. In fact you made me consider another case where response has not been received. I will make it even safer.

geek-soft: Now we will have:

```
            if (err.isAxiosError && err.response != null) {
                switch (err.response.status) {
```

src/modules/file-transfer-service/FileTransferService.ts
geek-soft: Well this is not in my hands, this is how API behaves: https://github.com/companieshouse/file-transfer-api/blob/master/src/routes/file-download.ts#L16

I basically went through an API code and reflected all scenarios in the service code.

n-r-b: ok, fair enough 👍

src/modules/file-transfer-service/FileTransferService.ts
n-r-b: i like the extra safety!

PR #135, opened 2020-04-09
feat: download just uploaded documents

akaJSaunders:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1589

### Change description

- add download links to list of files on the file upload page (pointing to /data/:fileId/download)
- check whether file is scanned and display error message if not it can be done either by handling FileNotReadyError or checking whether av_statusis equal scanned

### Work checklist

### Merge instructions

```
```


test/controllers/EvidenceUploadController.test.ts
```diff
@@ -98,6 +100,18 @@ describe('EvidenceUploadController', () => {
                         .and.to.contain('another-file.jpeg');
                 });
         });
+
+        it('should return 200 and render download links when attachments in session', async () => {
```
geek-soft, line 104: Scenario tested by this test is same as scenario tested by one above. I suggest merging the two together so one test checks both that filename is rendered on the screen and that it is linked to correct file download path.

test/controllers/EvidenceDownloadController.test.ts
```diff
@@ -26,15 +27,19 @@ describe('EvidenceDownloadController', () => {
     const contentLength = 1000;
     const contentType = `application/json`;
 
-    const metadata: FileMetadata = {
+    const expectedDownloadErrorHeading = 'The file can not be downloaded at this moment';
+    const expectedDownloadErrorMessage = 'Please try again later';
+
+    const createMetadata: FileMetadata = {
```
geek-soft, line 33: Name `createMetadata` sounds like function while in fact is is just object literal. Please rename so that is does not sound like function.

src/controllers/EvidenceDownloadController.ts
```diff
@@ -25,13 +29,32 @@ export class EvidenceDownloadController extends BaseAsyncHttpController {
...
+        const metadata: FileMetadata = await this.fileTransferService.getFileMetadata(fileId);
+
         res.setHeader('content-disposition', `attachment; filename=${metadata.name}`);
 
-        const readable = await this.fileTransferService.download(fileId);
+        if (!metadata.av_status || metadata.av_status !== 'clean') {
+            return this.renderDownloadError();
+        }
 
+        let readable = {} as Readable;
```
geek-soft, line 40: Having `{} as Readable` makes no sense. There is nothing in common between `{}` and `Readable`. Please remove value initiation.

test/controllers/EvidenceDownloadController.test.ts
```diff
@@ -109,4 +114,46 @@ describe('EvidenceDownloadController', () => {
...
+        const app = createDefaultApp(fileTransferServiceProxy(Promise.reject(
+            new FileNotReadyError(`File download failed because "${FILE_ID}" file is either infected or has not been scanned yet`)),
+            metadataClean));
+
+        await request(app)
+            .get(EXPECTED_DOWNLOAD_LINK_URL)
+            .then(res => {
+                expect(res.status).to.equal(FORBIDDEN);
+                expect(res.text).to.contain(expectedDownloadErrorHeading);
+                expect(res.text).to.contain(expectedDownloadErrorMessage);
```
geek-soft, line 129: May I ask you to use `.and` instead of having two expect calls?

test/controllers/EvidenceDownloadController.test.ts
```diff
@@ -109,4 +114,46 @@ describe('EvidenceDownloadController', () => {
...
+            return {
+                av_status: status as any,
+                content_type: contentType,
+                id: FILE_ID,
+                name: 'hello.txt',
+                size: contentLength
+            };
+        };
+
+        for (const status of [undefined, null, 'infected', 'not-scanned']) {
```
geek-soft, line 145: I think we can easily remove cases where status is undefined and null. These should never happen as per API contract.

n-r-b: ok, i've removed the undefined/null. This test actually helped me identify a small bug I had introduced to the controller :D

PR #142, opened 2020-04-10
refactor: AppealsStorageService is now standalone module called AppealsService

franciscog-kainos:
### JIRA link
N/A

### Change description
Renamed AppealsStorageService to AppealsService as it now supports more than just storing.

- Method to retrieve appeal by id.

### Work checklist

### Merge instructions

```
```


test/modules/appeals-service/AppealsService.test.ts
```diff
@@ -0,0 +1,233 @@
...
+        it('should return an appeal when valid arguments are provided', async () => {
+
+            nock(HOST)
+                .get(`${APPEALS_URI}/${APPEAL_ID}`)
+                .reply(200, appeal);
+
+            const returnedAppeal = await appealsService
+                .getAppeal(appeal.penaltyIdentifier.companyNumber, APPEAL_ID, BEARER_TOKEN);
+
+            expect(returnedAppeal).to.deep.eq(appeal);
```
geek-soft, line 194: May I suggest that in the future we use full form of equal method which is `.equal` which is bit more descriptive?

PR #152, opened 2020-04-14
feat: added middleware to load appeal from mongo

franciscog-kainos:
### JIRA link
Part of https://companieshouse.atlassian.net/browse/LFA-1467

### Change description
Added the middleware and tests.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/middleware/LoadAppealMiddleware.ts
+++ b/src/middleware/LoadAppealMiddleware.ts
@@ -9,7 +9,9 @@ import { BaseMiddleware } from 'inversify-express-utils';
 import { loggerInstance } from './Logger';
 
 import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
+import { companyNumberSchema } from 'app/models/PenaltyIdentifier.schema';
 import { AppealsService } from 'app/modules/appeals-service/AppealsService';
+import { SchemaValidator } from 'app/utils/validation/SchemaValidator';
 export const APPEAL_ID_QUERY_KEY = 'a';
 export const COMPANY_NUMBER_QUERY_KEY = 'c';
 
@@ -22,18 +24,24 @@ export class LoadAppealMiddleware extends BaseMiddleware {
     // @ts-ignore
     public async handler(req: Request, res: Response, next: NextFunction): Promise<void> {
 
+        const companyNumber = req.query[COMPANY_NUMBER_QUERY_KEY] as string;
+        const appealId = req.query[APPEAL_ID_QUERY_KEY] as string;
+
         const applicationData = req.session.chain(_ => _.getExtraData())
             .chainNullable<ApplicationData>(extraData => extraData[APPLICATION_DATA_KEY])
             .ifNothing(() => loggerInstance().error(`${LoadAppealMiddleware.name} - Could not retrieve application data from session`))
             .unsafeCoerce();
 
+        try {
+            new SchemaValidator(companyNumberSchema).validate(companyNumber);
+        } catch (err) {
+            throw Error('Tried to load appeal from an invalid company number');
+
+        }
         if (applicationData?.appeal) {
             loggerInstance().debug(`${LoadAppealMiddleware.name} - handler: user's session contains appeal`);
             return next();
         }
-
-        const companyNumber = req.query[COMPANY_NUMBER_QUERY_KEY] as string;
-        const appealId = req.query[APPEAL_ID_QUERY_KEY] as string;
 
         const token = req.session
             .chain(_ => _.getValue<ISignInInfo>(SessionKey.SignInInfo))
--- a/src/middleware/LoadAppealMiddleware.ts
+++ b/src/middleware/LoadAppealMiddleware.ts
@@ -8,7 +8,6 @@ import { provide } from 'inversify-binding-decorators';
 import { BaseMiddleware } from 'inversify-express-utils';
 import { loggerInstance } from './Logger';
 
-import { Appeal } from 'app/models/Appeal';
 import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
 import { AppealsService } from 'app/modules/appeals-service/AppealsService';
 export const APPEAL_ID_QUERY_KEY = 'a';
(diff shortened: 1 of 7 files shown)


src/modules/appeals-service/errors.ts
```diff
@@ -1,8 +1,29 @@
...
+
+interface StatusCode {
+    status: number;
+}
 // tslint:disable: max-classes-per-file
-export class AppealNotFoundError extends Error {}
 
-export class AppealServiceError extends Error {}
+export class AppealServiceError extends Error implements StatusCode {
+    public readonly status: number;
```
geek-soft, line 9: It does not make much sense to have status codes on service errors. It is down to app to interpret errors and react accordingly. Service should hide HTTP concerns from API consumers.

src/middleware/LoadAppealMiddleware.ts
```diff
@@ -0,0 +1,81 @@
...
+                .unsafeCoerce();
+
+            if (!appeal && appealId) {
+
+                try {
+                    appeal = await this.appealsService.getAppeal(companyNumber, appealId, token);
+                    this.transferAppealIntoSession(req, appeal);
+                } catch (err) {
+                    loggerInstance().error(`${LoadAppealMiddleware.name} - ${err.message} - ${JSON.stringify(err)}`);
+                    res.status(err.status);
```
geek-soft, line 54: Cause we have no specific error handling requirement at this point everything is treated as an unexpected error. Due to that it should have been thrown by middleware for handling in generic error handler that is designed to catch any unexpected errors.

src/middleware/LoadAppealMiddleware.ts
```diff
@@ -0,0 +1,81 @@
...
+                    res.render('error');
+                }
+
+            }
+        } catch (err) {
```
geek-soft, line 59: What trigger this scenario? Is that just `.unsafeCoerce()`? If so it should have been handled by generic error handler. Also if this is only case then `err.status` does not make much sense unless purifyTs sets value for status property.

src/modules/appeals-service/AppealsService.ts
```diff
@@ -21,7 +21,7 @@ export class AppealsService {
         loggerInstance()
             .debug(`Making a GET request to ${uri}`);
 
-        return await axios
+        return axios
```
geek-soft, line 24: Good spot :)

test/models/AppDataFactory.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const DEFALT_REASONS_WITH_ATTACHMENTS: Reasons = {
+    other: {
+        title: 'I have reasons',
+        description: 'They are legit',
+        attachments: DEFAULT_ATTACHMENTS
+    }
+};
+
+export const createAppData =
```
geek-soft, line 44: I think it adds virtually no value to me compared to what object literals can do. Fact it takes appeal and sets some default values when certain properties are missing is just magic that we do not need. Overall risks are grater than benefits.

test/models/AppDataFactory.ts
```diff
@@ -0,0 +1,73 @@
...
+        return {
+            appeal: {
+                penaltyIdentifier: appeal.penaltyIdentifier ? appeal.penaltyIdentifier : DEFAULT_PENALTY_IDENTIFIER,
+                reasons: appeal.reasons ? appeal.reasons : DEFALT_REASONS
+            },
+            navigation
+        };
+    };
+
+export const createDefaultAppDataWithAttachents = (navigation?: Navigation): Partial<ApplicationData> => {
```
geek-soft, line 55: Similarly here, method call can be easily replaced by:

```
{ appeal: APPEAL, navigation }
```

I find above more readable than seeing `createDefaultAppDataWithAttachents(navigation)`.

src/middleware/LoadAppealMiddleware.ts
```diff
@@ -0,0 +1,81 @@
...
+        const appealId = req.query[APPEAL_ID_QUERY_KEY] as string;
+
+        try {
+            const token = req.session
+                .chain(_ => _.getValue<ISignInInfo>(SessionKey.SignInInfo))
+                .chainNullable(signInInfo => signInInfo[SignInInfoKeys.AccessToken])
+                .chainNullable(accessToken => accessToken[AccessTokenKeys.AccessToken])
+                .unsafeCoerce();
+
+            if (!appeal && appealId) {
```
n-r-b, line 47: do we need to check if `!appeal`? as we are checking further up if we have an appeal

src/middleware/LoadAppealMiddleware.ts
```diff
@@ -0,0 +1,81 @@
...
+export const APPEAL_ID_QUERY_KEY = 'a';
+export const COMPANY_NUMBER_QUERY_KEY = 'c';
+
+@provide(LoadAppealMiddleware)
+export class LoadAppealMiddleware extends BaseMiddleware {
+
+    constructor(@inject(AppealsService) private readonly appealsService: AppealsService) {
+        super();
+    }
+    // @ts-ignore
```
n-r-b, line 23: is this needed?

src/middleware/LoadAppealMiddleware.ts
```diff
@@ -0,0 +1,81 @@
...
+                .chain(_ => _.getValue<ISignInInfo>(SessionKey.SignInInfo))
+                .chainNullable(signInInfo => signInInfo[SignInInfoKeys.AccessToken])
+                .chainNullable(accessToken => accessToken[AccessTokenKeys.AccessToken])
+                .unsafeCoerce();
+
+            if (!appeal && appealId) {
+
+                try {
+                    appeal = await this.appealsService.getAppeal(companyNumber, appealId, token);
+                    this.transferAppealIntoSession(req, appeal);
```
geek-soft, line 51: This `this.transferAppealIntoSession` way more complicated than it should have been. It would be simpler to do extract whole application data at line 26 and then do simple value assignment here `applicationData.appeal = appeal`.

Here is pseudo code:

```
let applicationData = req.session.chain(
....
if (applicationData.appeal) {
  return next();
}
...
applicationData.appeal = await this.appealsService.getAppeal(...
...
```

src/middleware/LoadAppealMiddleware.ts
geek-soft, line 47: No reason to check `!appeal` again since few lines higher we have `if (appeal) {` condition.

src/middleware/LoadAppealMiddleware.ts
franciscog-kainos: Nope, nice spot

test/models/AppDataFactory.ts
franciscog-kainos: I do not understand. As in APPEAL is a default appeal?

src/middleware/LoadAppealMiddleware.ts
```diff
@@ -0,0 +1,53 @@
...
+        const appealId = req.query[APPEAL_ID_QUERY_KEY] as string;
+
+        const token = req.session
+            .chain(_ => _.getValue<ISignInInfo>(SessionKey.SignInInfo))
+            .chainNullable(signInInfo => signInInfo[SignInInfoKeys.AccessToken])
+            .chainNullable(accessToken => accessToken[AccessTokenKeys.AccessToken])
+            .ifNothing(() => loggerInstance().error(`${LoadAppealMiddleware.name} - Could not retrieve token from session`))
+            .unsafeCoerce();
+
+        if (appealId) {
```
geek-soft, line 45: Shouldn't we check that company number is there as well? I guess it does not make sense to call API if either appeal ID or company number is empty. Other scenarios such as null or undefined are already caught in the service method.

test/middleware/LoadAppealMiddleware.test.ts
```diff
@@ -0,0 +1,122 @@
...
+import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
+import { AppealsService } from 'app/modules/appeals-service/AppealsService';
+import { AppealNotFoundError, AppealServiceError } from 'app/modules/appeals-service/errors';
+
+describe('LoadAppealMiddleware', () => {
+
+    const appealId = '123';
+    const companyId = 'NI000000';
+    // @ts-ignore
+    const createAppealService = (method: 'resolves' | 'rejects', data?: Appeal | any) => {
```
geek-soft, line 23: I do not see much value in having this method since there is `createSubstituteOf` ready for use as:

```
createSubstituteOf<Response>(service => {
    service.getAppeal(Arg.all()).resolves(data);
})
```

franciscog-kainos: sure I can use the `createSubstituteOf` method but at the end of the day I believe it saves a few lines and makes the code more readable.

```
const createAppealService = (method: 'resolves' | 'rejects', data?: Appeal | any) =>
        createSubstituteOf<AppealsService>(service => service.getAppeal(Arg.all())[method](data));

const service = createAppealService('resolves', something); // x4 vs.

const service = createSubstituteOf<AppealsService>(service => service.getAppeal(Arg.all()).resolves(something)); // x4
```

src/middleware/LoadAppealMiddleware.ts
franciscog-kainos: We probably should as we don't really check the company number on the backend at this stage

test/middleware/LoadAppealMiddleware.test.ts
geek-soft: General speaking number of lines is not decisive factor at all. Readability and maintainability is much more important. Having 5, 10, 15, 50 lines does not matter much to me. I look for the code which is simple, easy to read / follow and change.

PR #173, opened 2020-04-22
feat: make configuration profiles optional

geek-soft:
### JIRA link

None

### Change description

With this change application won't throw an error when configuration profile file is missing. In that case, cause file does not exist, environment variables won't be loaded and application will fail later during launch phase only if required configuration is missing.

I used an occasion to cleanup code and restore correct responsibilities split.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -1,0 +1,1 @@
+export const APP_NAME = 'lfp-appeals-frontend';
--- a/src/utils/ConfigLoader.ts
+++ b/src/utils/ConfigLoader.ts
@@ -1,13 +1,4 @@
-import bodyParser = require('body-parser');
-import { createLoggerMiddleware } from 'ch-logging';
-import cookieParser = require('cookie-parser');
 import * as dotenv from 'dotenv';
-import * as express from 'express';
-import * as nunjucks from 'nunjucks';
-import * as path from 'path';
-
-import { getEnv, getEnvOrThrow } from 'app/utils/EnvironmentUtils';
-import * as Paths from 'app/utils/Paths';
 
 const DEFAULT_ENV_FILE = `${__dirname}/../../.env`;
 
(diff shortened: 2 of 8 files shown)


src/Constants.ts
```diff
@@ -0,0 +1 @@
+export const APP_NAME = 'lfp-appeals-frontend';
```
akaJSaunders, line 1: Are we planning on using this Constants.ts in the future? Worried it might lead to some bad practices in the future where we just dump consts that we dont know where to put in it.  Could we maybe move this constant to another file or perhaps name this one specifically for the APP_NAME?

geek-soft: I have no defined plan for it right now, partially cause I do not know what the future is. 

Constants file was created cause it is better place than ConfigLoader. That is it.

src/utils/ConfigLoader.ts
```diff
@@ -1,71 +1,11 @@
...
-
-    app.locals.cdn = {
-        host: getEnvOrThrow('CDN_HOST')
-    };
-
-    const url = getEnv('PIWIK_URL');
-    const site = getEnv('PIWIK_SITE_ID');
-    if (url && site) {
-        app.locals.piwik = { url, site };
+        dotenv.config({ path: envFilePath });
```
franciscog-kainos, line 9: Do we want to log if the env.<NODE_ENV> file was not found?

geek-soft: I thought about it and concluded that there is more harm in displaying useless log entries than not having them at all.

At some point configuration profiles became optional and right now they are used only in local environments. In cloud environments environment variables are loaded outside of the application.

src/Constants.ts
franciscog-kainos: Alternatively it can be in `App.ts` or `Server.ts`

PR #179, opened 2020-04-27
Fix: lowercase legacy numbers

akaJSaunders:
### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/PenaltyDetailsController.ts
+++ b/src/controllers/PenaltyDetailsController.ts
@@ -25,11 +25,13 @@ const navigation = {
 
 const sanitizeForm = (body: PenaltyIdentifier) => {
     const legacyPrefix: string = 'PEN';
-    const penalty = body.penaltyReference.toUpperCase();
+    const penalty: string  = body.penaltyReference;
 
     return {
         companyNumber: sanitizeCompany(body.companyNumber),
-        penaltyReference: penalty.startsWith(legacyPrefix) ? sanitizeLegacyPenalty(penalty) : penalty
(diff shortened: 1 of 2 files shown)


src/controllers/PenaltyDetailsController.ts
```diff
@@ -24,12 +24,12 @@ const navigation = {
 };
 
 const sanitizeForm = (body: PenaltyIdentifier) => {
-    const legacyPrefixChars: string = 'PEN';
-    const penalty = body.penaltyReference;
+    const legacyPrefix: string = 'PEN';
+    const penalty = body.penaltyReference.toUpperCase();
```
n-r-b, line 28: i've noticed we are setting the penalty number to uppercase in sanitizeLegacyPenalty(), do we still need to do this?

n-r-b: Can you also add the type please

src/controllers/PenaltyDetailsController.ts
```diff
@@ -24,12 +24,12 @@ const navigation = {
...
 
     return {
         companyNumber: sanitizeCompany(body.companyNumber),
-        penaltyReference: penalty.startsWith(legacyPrefixChars) ? sanitizeLegacyPenalty(penalty) : penalty.toUpperCase()
```
franciscog-kainos, line 32: the `sanitizeLegacyPenalty` already upper cases the reference. You could have left this as it was or am I missing something?

n-r-b: i think he's doing this because `startsWith()` is case sensitive

PR #185, opened 2020-04-27
fix: log and return_url fix

franciscog-kainos:
### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/modules/appeals-service/AppealsService.ts
+++ b/src/modules/appeals-service/AppealsService.ts
@@ -43,7 +43,7 @@ export class AppealsService {
             .then((response: AxiosResponse<string>) => {
                 if (response.status === CREATED && response.headers.location) {
                     loggerInstance()
-                        .info(`${AppealsService.name} - save: created resource ${response.headers.location}`);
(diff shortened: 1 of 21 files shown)


src/modules/appeals-service/AppealsService.ts
```diff
@@ -43,7 +43,7 @@ export class AppealsService {
             .then((response: AxiosResponse<string>) => {
                 if (response.status === CREATED && response.headers.location) {
                     loggerInstance()
-                        .info(`${AppealsService.name} - save: created resource ${appeal.id} - ${response.headers.location}`);
+                        .info(`${AppealsService.name} - save: created resource ${response.data} - ${response.headers.location}`);
```
n-r-b, line 46: should this be r`esponse.data.toString()`?

franciscog-kainos: In this case it doesn't really matter because the log statement will always convert this to a string

franciscog-kainos: In the return statement it matters because we want to make sure that id that gets returned is a string and not a number for example

PR #192, opened 2020-05-04
fix: proxy for file transfer api

franciscog-kainos:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1699

### Change description
Set `proxy=false` in axios config for FileTransferService because it was automatically picking up the `http_proxy` variable in the vagrant environment. 

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/modules/file-transfer-service/FileTransferService.ts
+++ b/src/modules/file-transfer-service/FileTransferService.ts
@@ -34,15 +34,11 @@ export class FileTransferService {
         const data = new FormData();
         data.append('upload', file, fileName);
 
-        const config: AxiosRequestConfig = {
-            headers: {
+        return await axios
+            .post(this.url, data, this.getAxiosConfig({
                 ...this.prepareHeaders(),
                 ...data.getHeaders()
-            }
-        };
-
-        return await axios
-            .post(this.url, data, config)
+            }))
             .then((response: AxiosResponse) => { return response.data.id; })
             .catch(this.handleResponseError('upload', fileName));
     }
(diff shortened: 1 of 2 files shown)


src/modules/file-transfer-service/FileTransferService.ts
```diff
@@ -85,16 +75,24 @@ export class FileTransferService {
...
     }
 
+    private getAxiosConfig(headers: any = {}): AxiosRequestConfig {
+
+        return {
+            headers: {
+                ...this.prepareHeaders(),
+                ...headers
+            },
+            proxy: false
```
geek-soft, line 91: Isn't it possible to drive it via proxy variables? Isn't it possible to use NO_PROXY variable for it?

franciscog-kainos: In this occasion, what's the advantage in doing this?

geek-soft: Proxy is one of the things should be configurable not arbitrary set in software as settings differ from env to env.

franciscog-kainos: What do you suggest then? `constructor(private readonly url: string, private readonly key: string, useProxy: boolean)` something like this?

franciscog-kainos: Also responding to the first comment, I've researched on how it could be done and yes we could drive it using the NO_PROXY variable. By default it uses http_proxy

geek-soft: Perfect, in that case no code change is needed, am I right? Usually good HTTP libraries support proxy env out of the box, I hope it is the same case here.

franciscog-kainos: I'm going to test this now, but hopefully no code changes just some documentation around that to add

PR #238, opened 2020-06-15
feat: select year screen

franciscog-kainos:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1795

### Change description
Added select year screen
image

image

### Work checklist

### Merge instructions

```
```


src/views/select-the-year.njk
```diff
@@ -0,0 +1,57 @@
+{% extends 'layout.njk' %}
+
+{% from 'govuk/components/back-link/macro.njk'     import govukBackLink %}
```
akaJSaunders, line 3:
```suggestion
{% from 'govuk/components/back-link/macro.njk' import govukBackLink %}
```

src/views/select-the-year.njk
```diff
@@ -0,0 +1,57 @@
...
+{% from 'govuk/components/radios/macro.njk' import govukRadios %}
+
+{% block pageTitle %}
+  Select your penalty details
+{% endblock %}
+
+{% block backLink %}
+  {{ 
+    govukBackLink({
+      text: "Back",
```
akaJSaunders, line 14:
```suggestion
      text: 'Back',
```

src/controllers/PenaltyDetailsController.ts
```diff
@@ -19,7 +19,7 @@ const navigation = {
         return ROOT_URI;
     },
     next(): string {
-        return REVIEW_PENALTY_PAGE_URI;
+        return SELECT_YEAR_PAGE_URI;
```
akaJSaunders, line 22: ReviewPenaltyController.ts navigation needs to be updated due to this change

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,150 @@
...
+                                    ...that.prepareNavigationConfig()
+                                }
+                            );
+                        }
+                    }
+
+                    if (that.formSanitizeFunction != null) {
+                        request.body = that.formSanitizeFunction(request.body);
+                        loggerInstance().debug(`${BaseController.name} - sanitized form body: ${JSON.stringify(request.body)}`);
+                    }
```
akaJSaunders, line 119: I dont think you need these in your custom action handler

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,150 @@
...
+        const penaltyList: PenaltyList | undefined = appeal.penaltyIdentifier.penaltyList;
+
+        if (!penaltyList || !penaltyList.items) {
+            throw new Error(SelectYearController.PENALTY_EXPECTED_ERROR);
+        }
+
+        return { penalties: penaltyList.items.map(createPenaltyRadioButton) };
+    }
+
+    protected getExtraActionHandlers(): Record<string, FormActionHandler | FormActionHandlerConstructor> {
```
akaJSaunders, line 96: Not sure I understand why you needed a custom action handler

src/utils/Paths.ts
```diff
@@ -10,6 +10,7 @@ export const HEALTH_CHECK_URI = `${ROOT_URI}/healthcheck`;
 // Pages
 export const ENTRY_PAGE_URI = `${ROOT_URI}/start`;
 export const PENALTY_DETAILS_PAGE_URI = `${ROOT_URI}/penalty-reference`;
+export const SELECT_YEAR_PAGE_URI = `${ROOT_URI}/select-the-year`;
```
n-r-b, line 13: should the path be `/select-the-penalty`?

src/controllers/SelectYearController.ts
franciscog-kainos: image

Base Controller:
image

src/utils/Paths.ts
franciscog-kainos: yes oops

src/controllers/SelectYearController.ts
akaJSaunders, line 119: Are you doing any form of santization? I think we cant remove this

src/controllers/SelectYearController.ts
franciscog-kainos: If I don't include the view model when there is an error, the radio buttons will not populate with data because the assumption was made that the request body will contain the necessary data. Unfortunately when you submit the radio button form you only get the data for the selected radio button

src/views/select-the-year.njk
```diff
@@ -0,0 +1,57 @@
...
+      href: navigation.back.href
+    }) 
+  }}
+{% endblock %}
+
+{% block content %}
+
```
akaJSaunders, line 21:
```suggestion
```

src/utils/Paths.ts
n-r-b: please also update the const name

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,150 @@
...
+const template = 'select-the-year';
+
+const navigation = {
+    previous(): string {
+        return PENALTY_DETAILS_PAGE_URI;
+    },
+    next(): string {
+        return REVIEW_PENALTY_PAGE_URI;
+    },
+    actions: (changeMode: boolean) => {
```
n-r-b, line 33: Currently there is no option to change the penalty details from the summary screen, is this needed?

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,150 @@
...
+import { PenaltyReferenceRouter } from 'app/middleware/PenaltyReferenceRouter';
+import { Appeal } from 'app/models/Appeal';
+import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
+import { createPenaltyRadioButton } from 'app/models/components/PenaltyRadioButton';
+import { createSchema } from 'app/models/fields/PenaltyChoice.schema';
+import { APPLICATION_DATA_UNDEFINED, SESSION_NOT_FOUND_ERROR } from 'app/utils/CommonErrors';
+import { PENALTY_DETAILS_PAGE_URI, REVIEW_PENALTY_PAGE_URI, SELECT_YEAR_PAGE_URI } from 'app/utils/Paths';
+import { ValidationResult } from 'app/utils/validation/ValidationResult';
+
+const template = 'select-the-year';
```
n-r-b, line 24: please rename this to `select-the-penalty`

src/controllers/SelectYearController.ts
franciscog-kainos: image

No you're right however, I used the actions field for consistency as it is used in other action yielding controllers. I can make it indifferent to the change mode

src/views/select-the-year.njk
```diff
@@ -0,0 +1,57 @@
...
+  <div class="govuk-width-container">
+    <main role="main" class="govuk-main-wrapper">
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <form action="?{{ navigation.actions.continue }}" method="post" id="penaltyForm">
+
+            {{ 
+              govukRadios({
+                idPrefix: 'select-penalty',
+                name: 'selectPenalty',
```
n-r-b, line 31: this could also be `'select-penalty'`

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,150 @@
...
+            throw SESSION_NOT_FOUND_ERROR;
+        }
+
+        const appData: ApplicationData | undefined = session.getExtraData(APPLICATION_DATA_KEY);
+
+        if(!appData) {
+            throw APPLICATION_DATA_UNDEFINED;
+        }
+
+        const penaltySelected: string = request.body.selectPenalty;
```
n-r-b, line 56: this variable is redundant, could just do:
`appData.appeal.penaltyIdentifier.penaltyReference = request.body.selectPenalty;`

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,150 @@
...
+
+        appData.appeal.penaltyIdentifier.penaltyReference = penaltySelected;
+
+        session.setExtraData(APPLICATION_DATA_KEY, appData);
+    }
+
+}
+
+export const errorMessage: string = 'Select the penalty you want to appeal';
```
n-r-b, line 65: why not just add this to the `PenaltyChoice.schema.ts` since it is not being re-used elsewhere?

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,150 @@
...
+@controller(SELECT_YEAR_PAGE_URI, SessionMiddleware, AuthMiddleware, PenaltyReferenceRouter)
+export class SelectYearController extends SafeNavigationBaseController<any> {
+
+    public static PENALTY_EXPECTED_ERROR: string = 'Penalty object expected but none found';
+
+    constructor() {
+        super(
+            template,
+            navigation,
+            new FormValidator(Joi.object({
```
n-r-b, line 77: i feel it would be cleaner if we extract the form schema into it's own function

src/views/select-the-year.njk
franciscog-kainos: The name has to stay as is because that's the key name that gets returned on submit. 'select-penalty' is not a good key name in TS

src/models/components/PenaltyRadioButton.ts
```diff
@@ -0,0 +1,42 @@
...
+    label: {
+        classes: 'govuk-label--s';
+    },
+    hint: {
+        html: string;
+    };
+
+};
+
+export type PenaltyDetailsRadioComponent = {
```
n-r-b, line 15: i don't think this is used

src/controllers/SelectYearController.ts
akaJSaunders: thanks for the explanation

test/middleware/PenaltyReferenceRouter.test.ts
```diff
@@ -0,0 +1,184 @@
...
+import { createSession } from '../utils/session/SessionFactory';
+
+import { PenaltyReferenceRouter } from 'app/middleware/PenaltyReferenceRouter';
+import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
+import { Reasons } from 'app/models/Reasons';
+import { APPLICATION_DATA_UNDEFINED, SESSION_NOT_FOUND_ERROR } from 'app/utils/CommonErrors';
+import { REVIEW_PENALTY_PAGE_URI } from 'app/utils/Paths';
+
+
+
```
akaJSaunders, line 14:
```suggestion
```

akaJSaunders, line 14:
```suggestion

```

test/middleware/PenaltyReferenceRouter.test.ts
```diff
@@ -0,0 +1,184 @@
...
+import { createSubstituteOf } from '../SubstituteFactory';
+import { createSession } from '../utils/session/SessionFactory';
+
+import { PenaltyReferenceRouter } from 'app/middleware/PenaltyReferenceRouter';
+import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
+import { Reasons } from 'app/models/Reasons';
+import { APPLICATION_DATA_UNDEFINED, SESSION_NOT_FOUND_ERROR } from 'app/utils/CommonErrors';
+import { REVIEW_PENALTY_PAGE_URI } from 'app/utils/Paths';
+
+
```
akaJSaunders, line 13: No need for this large gap between tests and imports, also I noticed that you havnt put empty lines between tests, probably worth doing so.

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,137 @@
...
+                handle: async (request: Request, response: Response) => {
+                    if (that.validator != null) {
+                        const validationResult: ValidationResult = await that.validator.validate(request);
+                        if (validationResult.errors.length > 0) {
+                            return await that.renderWithStatus(UNPROCESSABLE_ENTITY)(
+                                that.template,
+                                {
+                                    ...request.body,
+                                    validationResult,
+                                    ...that.prepareViewModel(),
```
n-r-b, line 101: I'm wondering if other controllers would benefit from adding this line to the Base Controller's getDefaultActionHandler().. or can we not just add a custom validator that retains this data in some way?

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,137 @@
...
+        session.setExtraData(APPLICATION_DATA_KEY, appData);
+    }
+
+}
+
+// tslint:disable-next-line: max-classes-per-file
+@controller(SELECT_THE_PENALTY_PAGE_URI, SessionMiddleware, AuthMiddleware, PenaltyReferenceRouter)
+export class SelectYearController extends SafeNavigationBaseController<any> {
+
+    public static PENALTY_EXPECTED_ERROR: string = 'Penalty object expected but none found';
```
n-r-b, line 65: can this be `private`?

src/controllers/SelectYearController.ts
```diff
@@ -0,0 +1,137 @@
...
+        appData.appeal.penaltyIdentifier.penaltyReference = request.body.selectPenalty;
+
+        session.setExtraData(APPLICATION_DATA_KEY, appData);
+    }
+
+}
+
+// tslint:disable-next-line: max-classes-per-file
+@controller(SELECT_THE_PENALTY_PAGE_URI, SessionMiddleware, AuthMiddleware, PenaltyReferenceRouter)
+export class SelectYearController extends SafeNavigationBaseController<any> {
```
n-r-b, line 63: Can we rename this class to `SelectPenaltyController`?

src/controllers/SelectYearController.ts
franciscog-kainos: It's probably not a safe bet as some keys in the validation result can be the same as in the view model. Custom validator wouldn't really make it simpler as I would have to reconstruct the radio buttons in it

franciscog-kainos: If the default action handler was comprised of made of smaller reusable actions *hint* then this wouldn't be a problem

src/middleware/PenaltyReferenceRouter.ts
```diff
@@ -0,0 +1,50 @@
...
+            .appeal
+            .penaltyIdentifier
+            .penaltyList?.items
+            .length;
+
+        if (!numberOfPenalties) {
+            throw PenaltyReferenceRouter.PENALTY_LIST_UNDEFINED_ERROR;
+        }
+
+        if (numberOfPenalties === 1) {
```
n-r-b, line 38: can `numberOfPenalties` be 0?

src/models/PenaltyIdentifier.ts
```diff
@@ -2,7 +2,8 @@ import { PenaltyList } from 'ch-sdk-node/dist/services/lfp';
 
 export interface PenaltyIdentifier {
     companyNumber: string;
-    penaltyReference: string;
+    userInputPenaltyReference: string;
+    penaltyReference?: string;
```
n-r-b, line 6: should this be a required field? We could set this to the value of the `userInputPenaltyReference` so that it's never null

franciscog-kainos: It's not that it isn't required, it's that it won't be defined when the object is instantiated

src/middleware/PenaltyReferenceRouter.ts
franciscog-kainos: No but I'll change that to `<`

src/models/PenaltyIdentifier.ts
franciscog-kainos: There's some some logic behind this which does that after the user has pressed continue in the PenaltyDetailsPage

test/controllers/processors/AppealStorageFormSubmissionProcessor.test.ts
```diff
@@ -24,7 +24,7 @@ describe('AppealStorageForSubmissionProcessor', () => {
     const appeal: Appeal = {
         penaltyIdentifier: {
             companyNumber: '00345567',
-            penaltyReference: 'A00000001',
+            userInputPenaltyReference: 'A00000001',
```
n-r-b, line 27: what's the reason for changing this? only the `penaltyReference` is saved in mongo?

src/controllers/ReviewPenaltyController.ts
```diff
@@ -23,22 +27,44 @@ const navigation = {
...
+        const penaltyList = appeal.penaltyIdentifier.penaltyList;
+
+        if (!penaltyList) {
             throw new Error(ReviewPenaltyController.PENALTY_EXPECTED_ERROR);
+        }
+
+        let penaltyReference: string | undefined;
+
+        if (penaltyList.items.length === 1) {
+            penaltyReference = appeal.penaltyIdentifier.userInputPenaltyReference;
```
n-r-b, line 48: if there is only one item in the penaltyList, at which point does the `appeal.penaltyIdentifier.penaltyReference` get set?

franciscog-kainos: In the processor of the SelectPenaltyController

test/controllers/processors/AppealStorageFormSubmissionProcessor.test.ts
franciscog-kainos: My bad, didn't pay attention

src/controllers/SelectYearController.ts
n-r-b: I feel it would be cleaner if we created a custom validator over duplicating the default action handler, if it's not too much effort? We can reuse the logic we have in place to construct the radio buttons

src/controllers/ReviewPenaltyController.ts
n-r-b: what happens if the `PenaltyReferenceRouter` has redirected us to this page from the penalty details page?

franciscog-kainos: then it must mean that there is only one penalty reference and it will show you the table as it should? What are you asking?

n-r-b: i was just double checking that the `penaltyReference` is still stored in the session, as before

PR #241, opened 2020-06-18
feature: company prefix config

franciscog-kainos:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1632
https://companieshouse.atlassian.net/browse/LFA-1650

### Change description
Made company prefixes configurable.

### Work checklist

### Merge instructions

```
```


.env.vagrant
```diff
@@ -32,5 +32,7 @@ OAUTH2_TOKEN_URI=http://web.chs-dev.internal:4000/oauth2/token
 OAUTH2_CLIENT_ID=
 OAUTH2_CLIENT_SECRET=
 
+ALLOWED_COMPANY_PREFIXES=SC,NI,OC,SO,R,AP
```
akaJSaunders, line 35: Do we have a PR for updating all the default enviroment configs in the chs-config repo?

src/models/PenaltyIdentifierSchemaFactory.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    public readonly penaltyReferenceRegex: RegExp;
+
+    constructor(companyNumberPrefixes: string) {
+
+        if (companyNumberPrefixes.length === 0) {
+            throw new Error('Company prefixes must not be empty. e.g. NI,SC,SO');
+        }
+
+        if (companyNumberPrefixes === '*') {
```
akaJSaunders, line 18: What is meant by this? Isnt the list the one of the config? when would we have * as an item

.env.vagrant
franciscog-kainos: https://github.com/companieshouse/chs-configs/pull/1074/files

src/models/PenaltyIdentifierSchemaFactory.ts
franciscog-kainos: Making it configurable includes having a way to specify a way to match all company prefixes. `*` will match any company number between 1 and 8 chars long using  alphanumeric characters

.env.vagrant
n-r-b: I'm wondering whether it would be a good idea to add a note to the config ensuring 'AP' does not make it to the live environment?

n-r-b: could even have a separate property if we are worried this will make it into live

src/models/PenaltyIdentifierSchemaFactory.ts
```diff
@@ -0,0 +1,94 @@
...
+        } else {
+
+            const prefixesArray: string[] = companyNumberPrefixes.split(',');
+
+            const singleCharacterPrefixArray: string[] = prefixesArray.filter(prefix => prefix.length === 1);
+            const doubleCharacterPrefixArray: string[] = prefixesArray.filter(prefix => prefix.length === 2);
+
+            const prefixes: PrefixMap = {
+                singleCharacter: {
+                    lowerCase: this.makeLowerCase(singleCharacterPrefixArray),
```
akaJSaunders, line 29: Why do we need to lower case anything?

src/models/PenaltyIdentifierSchemaFactory.ts
```diff
@@ -0,0 +1,94 @@
...
+                    upperCase: this.makeUpperCase(singleCharacterPrefixArray)
+                },
+                doubleCharacter: {
+                    lowerCase: this.makeLowerCase(doubleCharacterPrefixArray),
+                    upperCase: this.makeUpperCase(doubleCharacterPrefixArray)
+                }
+            };
+
+            const singleCharacterRegex = `(${[
+                prefixes.singleCharacter.lowerCase.reduceRight((p, c) => `${p}|${c}`, '').substr(1),
```
akaJSaunders, line 39: Why do we get a substring when lowercase?

src/ContainerFactory.ts
```diff
@@ -73,6 +75,8 @@ export function createContainer(): Container {
 
     container.bind(CompaniesHouseSDK).toConstantValue(CompaniesHouseSDK(getEnvOrThrow('API_URL')));
 
+    container.bind(PenaltyIdentifierSchemaFactory).toConstantValue(new PenaltyIdentifierSchemaFactory(getEnvOrDefault('ALLOWED_COMPANY_PREFIXES', '*')));
```
n-r-b, line 78: can you split this over two lines for readability? it's quite a bit past the right hand margin in the IDE ;)

n-r-b: Also what does the default `*` give us? Could just throw an exception

src/models/PenaltyIdentifierSchemaFactory.ts
```diff
@@ -0,0 +1,94 @@
...
+            ].join('')})[0-9]{1,6}`;
+
+            const onlyNumbersRegex = '[0-9]{1,8}';
+
+            this.companyNumberRegex = new RegExp(`^(${
+                [singleCharacterRegex, doubleCharacterRegex, onlyNumbersRegex].join('|')
+                })$`);
+        }
+
+        this.penaltyReferenceRegex = /^[a-z0-9/]{8,14}$/i;
```
akaJSaunders, line 55: Like you mentioned on slack, do we need all this uppercase and lowercase logic considering we have the 'i' flag on the regex

akaJSaunders, line 55: we use the 'i' flag for the penaltyReference regex, could probably use it for the company number one and avoid all this uppercase and lowercase logic

src/models/PenaltyIdentifierSchemaFactory.ts
n-r-b: the ticket only mentions allowing company numbers with the prefix of SC, NI, OC, SO, R, so what would be the reason behind allowing anything outside of what is specified in this configuration?

src/models/PenaltyIdentifierSchemaFactory.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    }
+
+    public getCompanyNumberSchema(): Joi.StringSchema {
```
n-r-b, line 59: Are we still using the `PenaltyIdentifier.schema.ts` which contains the `companyNumberSchema`? What's the reason for duplicating what we currently have?

src/models/PenaltyIdentifierSchemaFactory.ts
akaJSaunders: ah its the default value when the list is empty? Shouldn't we just throw an error? Seems like a rather important configuration

src/models/PenaltyIdentifierSchemaFactory.ts
```diff
@@ -0,0 +1,94 @@
...
+            this.companyNumberRegex = /^[A-Za-z0-9]{8}$/i;
+        } else {
+
+            const prefixesArray: string[] = companyNumberPrefixes.split(',');
+
+            const singleCharacterPrefixArray: string[] = prefixesArray.filter(prefix => prefix.length === 1);
+            const doubleCharacterPrefixArray: string[] = prefixesArray.filter(prefix => prefix.length === 2);
+
+            const prefixes: PrefixMap = {
+                singleCharacter: {
```
n-r-b, line 28: Are company prefixes case sensitive? can we not just convert to uppercase?

.env.vagrant
akaJSaunders: Surely on the release notes we just have to list what should be on live.

src/models/PenaltyIdentifierSchemaFactory.ts
```diff
@@ -0,0 +1,94 @@
+import Joi from '@hapi/joi';
+
+type PrefixVariation = { upperCase: string[]; lowerCase: string[]; };
+type PrefixMap = { singleCharacter: PrefixVariation, doubleCharacter: PrefixVariation; };
+
+export class PenaltyIdentifierSchemaFactory {
```
n-r-b, line 6: The majority of the logic in this class is related to the company number, is it worth separating the two?

.env.vagrant
n-r-b: yep, that's fine as long as someone remembers this

src/models/PenaltyIdentifierSchemaFactory.ts
franciscog-kainos: No, I forgot to remove that file

src/models/PenaltyIdentifierSchemaFactory.ts
franciscog-kainos: I don't think it adds much value since throughout the project we tend to use the PenaltyIdentifier schema as a whole

src/models/PenaltyIdentifierSchemaFactory.ts
franciscog-kainos: I'll ask Bart to clarify that because the scenario 2 of my ticket is about making the prefixes configurable which in interpretation can mean including allowing any. Also, if for some reason the business decided to allow any prefix.

franciscog-kainos: image

franciscog-kainos: image

src/models/PenaltyIdentifierSchemaFactory.ts
n-r-b: ok fair enough

src/models/PenaltyIdentifierSchemaFactory.ts
```diff
@@ -11,46 +8,21 @@ export class PenaltyIdentifierSchemaFactory {
...
-                    upperCase: this.makeUpperCase(doubleCharacterPrefixArray)
-                }
-            };
+        const prefixesArray: string[] = companyNumberPrefixes.split(',');
 
-            const singleCharacterRegex = `(${[
-                prefixes.singleCharacter.lowerCase.reduceRight((p, c) => `${p}|${c}`, '').substr(1),
-                prefixes.singleCharacter.upperCase.reduceRight((p, c) => `${p}|${c}`, '')
-            ].join('')})[0-9]{1,7}`;
+        const prefixRegex = `(${prefixesArray.reduceRight((p, c) => `${p}|${c}`, '').substr(1)})[0-9]{1,7}`;
```
n-r-b, line 19: does this still work with both single and double prefix characters?

n-r-b: looking at the regex at the end: `[0-9]{1,7}`

franciscog-kainos: Yes you're right. Good spot

PR #244, opened 2020-06-22
feat: session cleanup after submission

franciscog-kainos:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1641

### Change description
After checking the penalty details the user is not able to go back to change them. The only permitted route is to go back to the beginning to the journey by pressing the govuk icon.
**Note:** The user can still access the confirmation screen 
image

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/CheckYourAppealController.ts
+++ b/src/controllers/CheckYourAppealController.ts
@@ -6,6 +6,7 @@ import { controller } from 'inversify-express-utils';
 import { SafeNavigationBaseController } from 'app/controllers/SafeNavigationBaseController';
 import { AppealStorageFormActionProcessor } from 'app/controllers/processors/AppealStorageFormActionProcessor';
 import { InternalEmailFormActionProcessor } from 'app/controllers/processors/InternalEmailFormActionProcessor';
+import { SessionCleanupProcessor } from 'app/controllers/processors/SessionCleanupProcessor';
 import { UserEmailFormActionProcessor } from 'app/controllers/processors/UserEmailFormActionProcessor';
 import { AuthMiddleware } from 'app/middleware/AuthMiddleware';
 import { loggerInstance } from 'app/middleware/Logger';
@@ -39,7 +40,7 @@ const navigation = {
 export class CheckYourAppealController extends SafeNavigationBaseController<any> {
     constructor() {
         super(template, navigation, undefined, undefined, [AppealStorageFormActionProcessor,
-            InternalEmailFormActionProcessor, UserEmailFormActionProcessor]);
+            InternalEmailFormActionProcessor, UserEmailFormActionProcessor, SessionCleanupProcessor]);
         // tslint:disable-next-line: forin
         for (const region in Region) {
             getEnvOrThrow(`${region}_TEAM_EMAIL`);
--- a/src/controllers/processors/SessionCleanupProcessor.ts
+++ b/src/controllers/processors/SessionCleanupProcessor.ts
@@ -3,6 +3,7 @@ import { Request } from 'express';
 import { provide } from 'inversify-binding-decorators';
 
 import { FormActionProcessor } from 'app/controllers/processors/FormActionProcessor';
+import { loggerInstance } from 'app/middleware/Logger';
 import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
 import { APPLICATION_DATA_UNDEFINED, SESSION_NOT_FOUND_ERROR } from 'app/utils/CommonErrors';
 import { CONFIRMATION_PAGE_URI } from 'app/utils/Paths';
@@ -31,6 +32,8 @@ export class SessionCleanupProcessor implements FormActionProcessor {
         appData.navigation.permissions = [CONFIRMATION_PAGE_URI];
 
         session.setExtraData(APPLICATION_DATA_KEY, appData);
+
+        loggerInstance().info(`${SessionCleanupProcessor.name} - Session data cleared`);
 
         return Promise.resolve();
     }
--- a/src/controllers/processors/SessionCleanupProcessor.ts
+++ b/src/controllers/processors/SessionCleanupProcessor.ts
@@ -1,0 +1,38 @@
+import { Session } from 'ch-node-session-handler';
+import { Request } from 'express';
+import { provide } from 'inversify-binding-decorators';
+
+import { FormActionProcessor } from 'app/controllers/processors/FormActionProcessor';
+import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
+import { APPLICATION_DATA_UNDEFINED, SESSION_NOT_FOUND_ERROR } from 'app/utils/CommonErrors';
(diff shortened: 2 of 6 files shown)


src/models/ApplicationData.ts
```diff
@@ -7,4 +7,5 @@ export const APPLICATION_DATA_KEY = 'appeals';
 export interface ApplicationData {
     appeal: Appeal;
     navigation: Navigation;
+    submittedAppeal?: Appeal;
```
akaJSaunders, line 10: Why store another appeal? Couldn't we just use a flag that returns true as suggested by the spike?

franciscog-kainos: The flag in itself would mean that we would have to somehow retrieve the appeal from the api. I took the best of both solutions described in the spike and implemented this instead. No required api calls and simple implementation

franciscog-kainos: Also, we are not storing another appeal, we are simply moving the submitted appeal to a 'confirmed' stage while we clear the 'build' stage for the new appeal

src/controllers/CheckYourAppealController.ts
```diff
@@ -39,7 +40,7 @@ const navigation = {
 export class CheckYourAppealController extends SafeNavigationBaseController<any> {
     constructor() {
         super(template, navigation, undefined, undefined, [AppealStorageFormActionProcessor,
-            InternalEmailFormActionProcessor, UserEmailFormActionProcessor]);
+            InternalEmailFormActionProcessor, UserEmailFormActionProcessor, SessionCleanupProcessor]);
```
akaJSaunders, line 43: why are we setting
`appData.submittedAppeal = appData.appeal`
at the check-your-appeal page? An appeal has not been submitted at this point

franciscog-kainos: At the point of execution of the SessionCleanupProcessor, which is after the appeal has been submitted by the AppealStorageFormActionProcessor, we can safely say that the appeal has been submitted to mongo and therefore we can mark this appeal as 'complete' by moving it to the `submittedAppeal` field. And deleting the current `appeal` from the session

src/controllers/processors/SessionCleanupProcessor.ts
```diff
@@ -0,0 +1,38 @@
+import { Session } from 'ch-node-session-handler';
+import { Request } from 'express';
+import { provide } from 'inversify-binding-decorators';
+
+import { FormActionProcessor } from 'app/controllers/processors/FormActionProcessor';
+import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
+import { APPLICATION_DATA_UNDEFINED, SESSION_NOT_FOUND_ERROR } from 'app/utils/CommonErrors';
+import { CONFIRMATION_PAGE_URI } from 'app/utils/Paths';
+
+@provide(SessionCleanupProcessor)
```
akaJSaunders, line 10: Maybe add one info/debug logging to inform that session has been cleaned up. What do you think?

src/controllers/CheckYourAppealController.ts
akaJSaunders: ah my bad, my head was in middleware mode, the processor makes more sense

test/controllers/CheckYourAppealController.test.ts
```diff
@@ -210,17 +241,22 @@ describe('CheckYourAppealController', () => {
                 'xUHinh19D17SQV2BYRLnGEZgeovYhcVitzLJMxpGxXW0w+30EYBb+6yF44pDWPsPejI17R5JSwy/Cw5kYQKO2A==';
 
             const appealsService = createSubstituteOf<AppealsService>(service => {
-                service.save(Arg.any(), Arg.any(), Arg.any()).returns(Promise.resolve(Arg.any()));
+                service.save(Arg.any(), Arg.any(), Arg.any()).returns(Promise.resolve('1'));
```
akaJSaunders, line 244: Why did you change this to 1?

src/controllers/processors/SessionCleanupProcessor.ts
```diff
@@ -0,0 +1,41 @@
...
+import { FormActionProcessor } from 'app/controllers/processors/FormActionProcessor';
+import { loggerInstance } from 'app/middleware/Logger';
+import { ApplicationData, APPLICATION_DATA_KEY } from 'app/models/ApplicationData';
+import { APPLICATION_DATA_UNDEFINED, SESSION_NOT_FOUND_ERROR } from 'app/utils/CommonErrors';
+import { CONFIRMATION_PAGE_URI } from 'app/utils/Paths';
+
+@provide(SessionCleanupProcessor)
+export class SessionCleanupProcessor implements FormActionProcessor {
+
+    public async process(req: Request): Promise<void> {
```
akaJSaunders, line 14: I think we need some unit tests for this processor

franciscog-kainos: I don't think we need that since all the logic was tested implicitly by the CheckYourAppealController.test.ts.

test/controllers/CheckYourAppealController.test.ts
franciscog-kainos: Substitute was having a hard time comparing Arg.any() when we do 
image

franciscog-kainos: The save method was saying that the args didn't match because { id: Arg.any() } did not equal { id: Arg.any() } :/

src/controllers/processors/SessionCleanupProcessor.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        if (!appData) {
+            throw APPLICATION_DATA_UNDEFINED;
+        }
+
+        appData.submittedAppeal = appData.appeal;
+
+        delete appData.appeal;
+
+        appData.navigation.permissions = [CONFIRMATION_PAGE_URI];
```
akaJSaunders, line 32: I understand why you are doing this, but what if a user submits an appeal, and then starts a new one, doesn't this mean they will have permissions to jump to the confirmation page at any time even though they might not have completed the new journey

franciscog-kainos: Yes they will be able to see the appeal they just submitted. Sounds like a feature to me.

franciscog-kainos: Harmless feature

src/controllers/CheckYourAppealController.ts
n-r-b: Let's say the customer has submitted an appeal, and they navigate back to the start of the journey, will their stale appeal data from their active session populate the fields on the `appeal-a-penalty/penalty-reference` page? Or does this get cleared before they get to the summary page?

src/controllers/processors/SessionCleanupProcessor.ts
akaJSaunders: True, I guess you are using the appealSubmitted data for that

PR #247, opened 2020-07-14
add: company authentication page

akaJSaunders:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1266

### Change description

Feature flag included

Test Company Number for Docker enviroment: SC109400 (use any legacy penalty reference)

Authorization code is 222222

### Work checklist

### Merge instructions

```
```


src/ContainerFactory.ts
```diff
@@ -78,6 +81,27 @@ export function createContainer(): Container {
...
 
+    const companyAuthConfig: CompanyAuthConfig = {
+        oath_scope_prefix: getEnvOrThrow('OATH_SCOPE_PREFIX'),
+        accountUrl: getEnvOrThrow('ACCOUNT_URL'),
+        accountRequestKey: getEnvOrThrow('OAUTH2_REQUEST_KEY'),
+        accountClientId: getEnvOrThrow('OAUTH2_CLIENT_ID'),
+        chsUrl: getEnvOrThrow('ACCOUNT_WEB_URL'),
+    };
+
+    const sessionStoreForAuthConfig = {
```
geek-soft, line 92: I suggest creating class for this config which has static method `createFromEnvironmentVariables()` which returns instance populated with values read from env variables. This way it can be used here and in base controller to remove duplicates and keeps defaults consistent.

Example usage:

```
const config = SessionConfig.createFromEnvironmentVariables();
```

src/models/CompanyAuthConfig.ts
```diff
@@ -0,0 +1,7 @@
+export default interface CompanyAuthConfig {
```
geek-soft, line 1: Why???

No default exports please.

src/models/sessionStoreConfig.ts
```diff
@@ -0,0 +1,6 @@
+export default interface SessionStoreConfig {
```
geek-soft, line 1: Why???

No default exports please.

test/ApplicationFactory.ts
```diff
@@ -39,6 +40,9 @@ export const createApp = (data?: Partial<ApplicationData>,
         const session: Session | undefined = data ? configureSession(createSession(cookieSecret)) : undefined;
         session?.setExtraData(APPLICATION_DATA_KEY, data);
 
+        // @ts-ignore
+        session?.data.signin_info.company_number = 'NI000000';
```
geek-soft, line 44: This is needed for certain cases only, why is that added here? I am thinking having this here make is buried so deep that it will be difficult to understand how company number is set. I'd prefer to see it set explicitly for tests that need it.

src/models/CompanyAuthConfig.ts
```diff
@@ -0,0 +1,7 @@
...
+    oath_scope_prefix: string;
+    accountUrl: string;
```
geek-soft, line 3: Do we need separate ACCOUNT URL and CHS URL? What isCHS URL needed for?

akaJSaunders: In the enviroment variables there are these two:
```
ACCOUNT_URL=http://account.chs.local
ACCOUNT_WEB_URL=http://chs.local
```
ACCOUNT_WEB_URL was already present and is the CHS_URL in my code. I have tried using ACCOUNT_WEB for both of those variables, but it broke the authorisation journey.

geek-soft: I need to understand more what CHS_URL is needed for.

src/modules/jwt-encryption-service/JwtEncryptionService.ts
```diff
@@ -0,0 +1,49 @@
...
+import { JWE, JWK } from 'node-jose';
+
+import CompanyAuthConfig from 'app/models/CompanyAuthConfig';
+
+interface AuthPayload {
+    nonce: string;
+    content: string;
+}
+
+export default class JwtEncryptionService {
```
geek-soft, line 13: Again, no default exports please.

src/modules/jwt-encryption-service/JwtEncryptionService.ts
```diff
@@ -0,0 +1,49 @@
...
+    public constructor(private companyAuthConfig: CompanyAuthConfig) {}
+
+    public generateNonce(): string {
+        const bytes = randomBytes(5);
+        const buffer = Buffer.from(bytes);
+        return buffer.toString('base64');
+    }
+
+    public async jweEncodeWithNonce(returnUri: string, nonce: string): Promise<string> {
```
geek-soft, line 22: Arg should not be called `returnUri` because service has nothing to do with URI's or redirects.

src/modules/jwt-encryption-service/JwtEncryptionService.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import CompanyAuthConfig from 'app/models/CompanyAuthConfig';
+
+interface AuthPayload {
+    nonce: string;
+    content: string;
+}
+
+export default class JwtEncryptionService {
+    public constructor(private companyAuthConfig: CompanyAuthConfig) {}
```
geek-soft, line 14: Config has 5 properties while only one is used, don't you think it is bad design? I would think it is better to keep that class generic as JWE encryption service which takes payload and key and returns encrypted payload.

src/modules/jwt-encryption-service/JwtEncryptionService.ts
geek-soft, line 22: I would consider rename as for me it is not encoding process, it is encrypting process.

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const service: JwtEncryptionService = new JwtEncryptionService({
+        oath_scope_prefix: '',
+        chsUrl: '',
+        accountClientId: '',
+        accountRequestKey: 'pXf+qkU6P6SAoY2lKW0FtKMS4PylaNA3pY2sUQxNFDk=',
+        accountUrl: ''
+    });
+
+    it('should generate random nonce value in base64', () => {
```
geek-soft, line 16: Not sure what that test does? It does not test randomness (because it is invoked only once). It does not test base 64 encoding really (because broken base 64 encoded string can still conform regex).

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        const nonce = service.generateNonce();
+        const test = /[A-Za-z0-9+/=]/.test(nonce);
+        assert.equal(nonce[nonce.length - 1], '=');
+        assert.equal(test, true);
+    });
+
+    it('should create an encrypted state string using nonce value', async () => {
+        const nonce = '2dsa=';
+        const state = await service.jweEncodeWithNonce('http://example.com', nonce);
+        const test = /eyJhbGciOiJkaXIiLCJlbmMiOiJBMTI4Q0JDLUhTMjU2Iiwia2lkIjoia2V5In0..*/.test(state);
```
geek-soft, line 26: What `eyJhbGciOiJkaXIiLCJlbmMiOiJBMTI4Q0JDLUhTMjU2Iiwia2lkIjoia2V5In0` is? Would't it better to decrypt it back and compare decrypted value?

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import JwtEncryptionService from 'app/modules/jwt-encryption-service/JwtEncryptionService';
+
+
+describe('JwtEncryptionService', () => {
+
+    const service: JwtEncryptionService = new JwtEncryptionService({
+        oath_scope_prefix: '',
+        chsUrl: '',
+        accountClientId: '',
+        accountRequestKey: 'pXf+qkU6P6SAoY2lKW0FtKMS4PylaNA3pY2sUQxNFDk=',
```
geek-soft, line 12: That only confirms bad design I highlighted earlier.

src/models/CompanyAuthConfig.ts
akaJSaunders: this is where they are both used:

```
return `${this.authConfig.accountUrl}/oauth2/authorise`.concat(
            '?',
            `client_id=${this.authConfig.accountClientId}`,
            `&redirect_uri=${this.authConfig.chsUrl}/oauth2/user/callback`,
            `&response_type=code`,
            `&scope=${scope}`,
            `&state=${encodedNonce}`);
```
accountURL serves as the domain of the URI we send the user to for the authorisation journey, and chsUrl is used for the callback parameter

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
akaJSaunders: How could this be tested? Running it multiple times and checking if values are different? two random values could happen to be equal (even tho chances are slim) and fail the test.

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
akaJSaunders: That would involve implementing a decode functionality for a test. What do you think?

test/ApplicationFactory.ts
akaJSaunders: I originally wanted to add it to the `createSession` of `SessionFactory.ts`, where all other information is added however that is type checked. The current location is just after the `createSession` is called.

I cant figure out how to get access to the session through the individual tests, as the session is enclosed in the `app` that is created by this factory.

src/ContainerFactory.ts
akaJSaunders: I have created the class and used it for both the base controller and the container factory

src/middleware/CompanyAuthMiddleware.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        try {
+            const uri = await this.getAuthRedirectUri(req, res, companyNumber);
+
+            loggerInstance().debug(`CompanyAuthMiddleware: Redirecting to ${uri}`);
+            return res.redirect(uri);
+
+        } catch (err){
+            console.log(err);
```
geek-soft, line 55: Remove, we do not log to console. It was only useful for debugging.

src/middleware/CompanyAuthMiddleware.ts
```diff
@@ -0,0 +1,107 @@
...
+            next(err);
+        }
+    }
+
+    async getAuthRedirectUri(req: Request, res: Response, companyNumber: string): Promise<string> {
+
+        const originalUrl: string = req.originalUrl;
+        const scope: string = this.authConfig.oathScopePrefix + companyNumber;
```
geek-soft, line 63: How is scope verified in the process? Does it matter? What would have happened if value provided by us was wrong?

test/ApplicationFactory.ts
geek-soft: What test needs that?

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
geek-soft: Can randomness be increased beyond 5 bytes? That touches on question that Rhys had.

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
geek-soft: I do not see anything wrong in implementing decrypt function for the test especially that most of that heavy lifting is delegated to libraries. Feel better than the test that has magic number like `eyJhbGciO...`

These tests here look more like tests that needed to be added for sake of having them. Tests should be useful not just exist.

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    };
+
+    const sessionStoreForAuthConfig = {
+        sessionCookieName: '__SID',
+        sessionCookieDomain: 'rebel1.aws.chdev.org',
+        sessionCookieSecureFlag: 'true',
+        sessionTimeToLiveInSeconds: 3600
+    };
+
+    const featureFlag: string = '1';
```
geek-soft, line 51: There is no test for value of 0. I'd think that having test for that making sure middleware is not triggered when feature is enabled is quite important.

.env.vagrant
```diff
@@ -29,9 +29,14 @@ CDN_HOST=chs-dev.internal:8080
 
 APPEALS_API_URL=http://api.chs-dev.internal
 
+OATH_SCOPE_PREFIX=https://api.companieshouse.gov.uk/company/
 OAUTH2_TOKEN_URI=http://web.chs-dev.internal:4000/oauth2/token
 OAUTH2_CLIENT_ID=
 OAUTH2_CLIENT_SECRET=
+OAUTH2_REQUEST_KEY=
```
geek-soft, line 36: Are OAUTH2_REQUEST_KEY, OATH_SCOPE_PREFIX and ACCOUNT_URL still needed here? I'd suggest to remove these unless needed otherwise they just cause confusion that you experienced last time.

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        );
+
+        await companyAuthMiddleware.handler(request, response, nextFunction);
+        nextFunction.didNotReceive();
+        response.received(1).redirect(Arg.any());
+
+    });
+});
+
+const createEncryptionService = (method: 'resolves' | 'rejects', encoded?: string | any)
```
geek-soft, line 130: Not a big fan of functions like that. Feels strange to have to say resolve / reject as via string arg. Here is how we used to do it before:

```
const emailService = createSubstituteOf<EmailService>(service => {
    service.send(Arg.any()).returns(Promise.reject(Error('Unexpected error')));
});
```

In above example it is clear to me what substitute of email service is setup to do.

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        );
+
+        const nextFunction = createSubstituteOf<NextFunction>();
+        const response = createSubstituteOf<Response>();
+        const request = { session: undefined } as Request;
+
+        try{
+            await companyAuthMiddleware.handler(request, response, nextFunction);
+        } catch (err) {
+            expect(err.message).to.equal(SESSION_NOT_FOUND_ERROR.message);
```
geek-soft, line 73: How does it work? Test says that next should be called while exception is being caught. I thought you will do verification of what arg next function was invoked with.

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -0,0 +1,155 @@
...
+            sessionStore,
+            encryptionService,
+            companyAuthConfig,
+            sessionStoreForAuthConfig,
+            featureFlag
+        );
+
+        await companyAuthMiddleware.handler(request, response, nextFunction);
+        nextFunction.didNotReceive();
+        response.received(1).redirect(Arg.any());
```
geek-soft, line 125: Is making any redirect a purpose of the company middleware? Saying middleware works fine when any redirect is made is over simplification. Don't you think?

test/ApplicationFactory.ts
akaJSaunders: Review Penalty Controller, the page that the middleware intercepts at the moment, if the `session?.data.signin_info.company_number` is not set, the middleware will redirect rather than let the page render

src/models/SessionConfig.ts
```diff
@@ -0,0 +1,22 @@
+import { getEnvOrDefault, getEnvOrThrow } from 'app/utils/EnvironmentUtils';
+
+export interface SessionStoreConfig {
+    sessionCookieName: string;
+    sessionCookieDomain: string;
+    sessionCookieSecureFlag: string;
+    sessionTimeToLiveInSeconds: number;
+}
+
+export class SessionConfig {
```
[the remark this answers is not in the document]
akaJSaunders: tslint expects `createFromEnviromentVariables()` to have a type definition, and the interface provided that. What if I change the naming of the interface and remove the export? what do you think?

src/middleware/CompanyAuthMiddleware.ts
akaJSaunders: This documentation explains the role of scope and their values quite well: https://companieshouse.atlassian.net/wiki/spaces/Arch/pages/1334771713/API+Filing+Service+Designs+3rd+Parties+and+CH

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
akaJSaunders: I increased it to 10, and it seems to work fine. Cant find documentation on any requirements.

src/middleware/CompanyAuthMiddleware.ts
akaJSaunders: The convention for OAuth scopes is to provide the full URI of the resource including the domain then, add a period followed by the permission level e.g. 

the scope “https://api.companieshouse.gov.uk/user/profile.read” would provide the requesting client access to READ the user profile resource.

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
geek-soft: Seems like nounce does not have to be too long, sometimes 32 bits (4 bytes) is enough and to be fairs as long as these are cryptographically random chance of collision is very small. Lets keep it at current length and lets generate it twice to make sure they are not the same.

src/models/SessionConfig.ts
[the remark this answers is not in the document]
geek-soft: Class provides type definitions as well. Why not to have a class with fields and static function?

.env.vagrant
```diff
@@ -33,6 +33,8 @@ OAUTH2_TOKEN_URI=http://web.chs-dev.internal:4000/oauth2/token
 OAUTH2_CLIENT_ID=
 OAUTH2_CLIENT_SECRET=
 
+COMPANY_AUTH_FEATURE_FLAG=0
```
geek-soft, line 36: May I suggest different pattern? Out API used `CHIPS_FEATURE_ENABLED` which is bit more readable to me. Why not to have `COMPANY_AUTH_FEATURE_ENABLED` or `COMPANY_AUTH_VERIFICATION_FEATURE_ENABLED`

Names should be as clear as possible cause they will be maintained by people outside of the project.

src/middleware/CompanyAuthMiddleware.ts
```diff
@@ -0,0 +1,105 @@
...
+                private readonly encryptionService: JwtEncryptionService,
+                private readonly authConfig: CompanyAuthConfig,
+                private readonly sessionStoreConfig: SessionStoreConfig,
+                private readonly featureFlag: boolean) {
+        super();
+    }
+
+    public handler: RequestHandler = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+
+        if(!this.featureFlag){
```
geek-soft, line 28: Reformat please.

src/middleware/CompanyAuthMiddleware.ts
geek-soft: What happens if you provide bad value, e.g. `foo bar` as a scope?

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    };
+
+    const sessionStoreForAuthConfig = {
+        sessionCookieName: '__SID',
+        sessionCookieDomain: 'rebel1.aws.chdev.org',
+        sessionCookieSecureFlag: 'true',
+        sessionTimeToLiveInSeconds: 3600
+    };
+
+    const redirectUrl: string = 'MOCK/oauth2/authorise?client_id=MOCK&redirect_uri=MOCK/oauth2/user/callback&response_type=code&scope=MOCKSC123123&state=';
```
geek-soft, line 51: Because you use MOCL for all elements of configuration it won't detect cases where you implemented bad string concatenation and for example `accountUrl` is used twice while `chsUrl` is never used.

Provide unique sample values for tests that will give us a chance to detect potential code issue.

src/modules/jwt-encryption-service/JwtEncryptionService.ts
```diff
@@ -0,0 +1,54 @@
...
+        return await JWE.createEncrypt({
+            format: 'compact',
+            fields: {
+                alg: 'dir',
+                enc: 'A128CBC-HS256'
+            }
+        }, key).update(payload).final();
+    }
+
+    public async jweDecryptWithNonce(content: string, requestKey: string): Promise<JWE.DecryptResult> {
```
geek-soft, line 38: That name does not make sense. There is no nounce. Just base64 encoded value is being decrypted with key.

src/modules/jwt-encryption-service/JwtEncryptionService.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export class JwtEncryptionService {
+
+    public generateNonce(): string {
+        const bytes = randomBytes(10);
+        const buffer = Buffer.from(bytes);
+        return buffer.toString('base64');
+    }
+
+    public async jweEncryptWithNonce(content: string, nonce: string, requestKey: string): Promise<string> {
```
geek-soft, line 14: For me it makes more sense to combine `content: string, nonce: string` into one argument of object type. It gives more flexibility for it to be used in more generic use cases. Nounce is not relevant here, it is just payload.

After removing nounce from arg list please also remove `WithNounce` from name.

I'd generally keep it `encrypt` and `decrypt` cause class name defines what encryption it is.

src/middleware/CompanyAuthMiddleware.ts
akaJSaunders: The scope is the concatenation of  OATH_SCOPE_PREFIX and companyNumber
If you set the environment variable OATH_SCOPE_PREFIX to an incorrect value, it will break when the user is intercepted by the middleware

test/modules/jwt-encryption-service/JwtEncryptionService.test.ts
```diff
@@ -0,0 +1,37 @@
...
+        assert.equal(nonce1[nonce1.length - 1], '=');
+        assert.equal(nonce2[nonce2.length - 1], '=');
+        assert.equal(test1, true);
+        assert.equal(test2, true);
+    });
+
+    it('should create an encrypted state and decrypt correctly', async () => {
+        const nonce = '2dsa=';
+        const content = 'http://example.com';
+        const requestKey = 'pXf+qkU6P6SAoY2lKW0FtKMS4PylaNA3pY2sUQxNFDk=';
```
geek-soft, line 28: Can you use other key than one used in our env? I am not sure how it is being used and I'd rather avoid leaking secret that might be sensitive.

PR #249, opened 2020-07-20
fix: Unconditional Company Authentication

rhysbarrett:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1265

### Change description

Company authentication now takes place _before_ the Select Penalty page is displayed (instead of afterwards, before the Review Penalty page), for additional security and code brevity. See Diagram:

IMG_20200720_115202

**Note**: Service cookies should be cleared before using this code for the first time.

### Work checklist

### Merge instructions

```
```


src/controllers/SelectPenaltyController.ts
```diff
@@ -55,7 +56,8 @@ class Processor implements FormActionProcessor {
 }
 
 // tslint:disable-next-line: max-classes-per-file
-@controller(SELECT_THE_PENALTY_PAGE_URI, SessionMiddleware, AuthMiddleware, PenaltyReferenceRouter)
+@controller(SELECT_THE_PENALTY_PAGE_URI, SessionMiddleware, AuthMiddleware, PenaltyReferenceRouter,
```
geek-soft, line 59: I do not think `CompanyAuthMiddleware` will be invoked when `PenaltyReferenceRouter` did redirect to `REVIEW_PENALTY_PAGE_URI`, although I might be always wrong.

Despite above we need at least few tests that make sure that:
- auth code is enforced when legacy penalty ref is used and only one penalty was found
- auth code is enforced when legacy penalty ref is used and more than one penalty was found
- no auth code is enforced for non legacy penalty ref

test/ApplicationFactory.ts
```diff
@@ -45,10 +48,16 @@ export const createApp = (data?: Partial<ApplicationData>,
...
         const signature = session?.data[SessionKey.ClientSig];
-
         const cookie = session ? Cookie.createFrom(sessionId! + signature) : null;
-
         const sessionStore = Substitute.for<SessionStore>();
+        const sessionConfig: SessionStoreConfig  = SessionStoreConfig.createFromEnvironmentVariables();
+        const encryptionService = new JwtEncryptionService();
+        const companyAuthConfig: CompanyAuthConfig = {
+            accountUrl: 'localhost',
+            accountRequestKey: 'pXf+qkU6P6SAoY2lKW0FtKMS4PylaNA3pY2sUQxNFDk=',
```
geek-soft, line 57: Lets's use different value. @akaJSaunders put a test where key is created by repeating 'a' number of times.

test/ApplicationFactory.ts
```diff
@@ -45,10 +48,16 @@ export const createApp = (data?: Partial<ApplicationData>,
...
         const sessionId = session?.data[SessionKey.Id];
         const signature = session?.data[SessionKey.ClientSig];
-
         const cookie = session ? Cookie.createFrom(sessionId! + signature) : null;
-
         const sessionStore = Substitute.for<SessionStore>();
+        const sessionConfig: SessionStoreConfig  = SessionStoreConfig.createFromEnvironmentVariables();
+        const encryptionService = new JwtEncryptionService();
+        const companyAuthConfig: CompanyAuthConfig = {
+            accountUrl: 'localhost',
```
geek-soft, line 56: Lets use 'http://account.chs' so we can distinguish it in test from chs url.

test/ApplicationFactory.ts
```diff
@@ -45,10 +48,16 @@ export const createApp = (data?: Partial<ApplicationData>,
...
+            accountRequestKey: 'pXf+qkU6P6SAoY2lKW0FtKMS4PylaNA3pY2sUQxNFDk=',
+            accountClientId: 'test',
+            chsUrl: 'test',
```
geek-soft, line 59: Lets use 'http://chs'

test/controllers/SelectPenaltyController.test.ts
```diff
@@ -40,13 +40,49 @@ describe('SelectPenaltyController', () => {
...
+        await request(app)
+            .get(SELECT_THE_PENALTY_PAGE_URI)
+            .expect(res => {
+                expect(res.text).to.include('type="radio"');
+                expect(res.text).to.include('value="A0000001"');
+                expect(res.text).to.include('value="A0000002"');
+            });
+    });
+
+    it('GET: should redirect to Company Authentication when user is not authenticated', async () => {
```
geek-soft, line 55: Didn't we agree to have separate test file which makes GET request to set of paths and expects to get that same redirect for every request made?

rhysbarrett: We did, but then when I attempted to implement this I realised that each controller we'd like to apply this iterative test approach to can have differing binding configurations when calling `createApp`. For example, compare [this line](https://github.com/companieshouse/lfp-appeals-frontend/pull/249/files#diff-a2e1a9f6667fe3d3541902b5f0766ff3R44) from Select Penalty with [this line](https://github.com/companieshouse/lfp-appeals-frontend/pull/249/files#diff-f810a152defbda778868a9b40bcf50beR87) from Review Penalty.

geek-soft: Does it really matter what data is in the session since company auth middleware is triggered before view rendering?

I would have thought that only thing middleware is concerned is:
- code session to be populated
- correct navigation entry to be present
- different company number to be in two places

rhysbarrett: It's more about the binding configuration than the session. But I understand what you mean - see updated test.

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -68,10 +73,10 @@ describe('Company Authentication Middleware', () => {
 
         const nextFunction = createSubstituteOf<NextFunction>();
         const response = createSubstituteOf<Response>();
-        const request = { session: undefined } as Request;
+        const req = { session: undefined } as Request;
```
geek-soft, line 76: That rename seems pointless, please revert.

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -156,13 +161,57 @@ describe('Company Authentication Middleware', () => {
...
 
-        await companyAuthMiddleware.handler(request, response, nextFunction);
+        await companyAuthMiddleware.handler(req, response, nextFunction);
         nextFunction.didNotReceive();
         response.received(1).redirect(Arg.is(startingWithRedirectUrl));
 
     });
 });
 
+describe('All pages after the Penalty Details page:', () => {
```
geek-soft, line 171: As we discussed please move it into its own test file.

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -156,13 +161,57 @@ describe('Company Authentication Middleware', () => {
...
+            const applicationData: Partial<ApplicationData> = {
+                appeal: {
+                    penaltyIdentifier: {
+                        companyNumber: 'NI999999',
+                        penaltyReference: 'PEN1A/NI000000',
+                        companyName: 'Test',
+                        penaltyList: {
+                            items: [
+                                {} as Penalty,
+                                {} as Penalty
```
geek-soft, line 190: I guess number of items does not matter now so is that even needed? Shall we remove company name and penalty list from session data?

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -156,13 +161,57 @@ describe('Company Authentication Middleware', () => {
...
+                            ]
+                        } as PenaltyList
+                    }
+                } as Appeal,
+                navigation: { permissions: [ page[1] ] }
+            };
+
+            /* TODO: Work out how to use this function call to override default Company Number
```
geek-soft, line 198: What are you planning with this? Will you just leave it there forever?

test/middleware/CompanyAuthMiddleware.test.ts
rhysbarrett: You're right. I tried removing the penalty list a while ago but the tests were failing so I left them in. Now that I've fixed the middleware order I'm able to remove the surplus session data.

test/middleware/CompanyAuthMiddleware.test.ts
rhysbarrett: Done

test/middleware/AppliedCompanyAuthMiddleware.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { Appeal } from 'app/models/Appeal';
+import { ApplicationData } from 'app/models/ApplicationData';
+import { REVIEW_PENALTY_PAGE_URI, SELECT_THE_PENALTY_PAGE_URI } from 'app/utils/Paths';
+
+
+describe('All pages after the Penalty Details page:', () => {
+
+    const pageList = [
+        ['Select Penalty', SELECT_THE_PENALTY_PAGE_URI],
```
geek-soft, line 14: Data structure is bit archaic. With destructuring assignment in TS / JS it could been an array of objects which you could destruct into named arguments in the test method.

Doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

rhysbarrett: Agreed, done.

PR #250, opened 2020-07-21
feat: apply company auth middleware to all pages in appeals journey

n-r-b:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1295

### Change description
Applied company auth middleware to most routes in the appeals journey (except from landing, penalty details and document download) so that the customer cannot bypass the company code authentication screen.

### Work checklist

### Merge instructions

```
```


test/controllers/CheckYourAppealController.test.ts
```diff
@@ -83,28 +83,16 @@ describe('CheckYourAppealController', () => {
                 });
         });
 
-        it('should return 200 with no populated session data', async () => {
+        it('should return 500 when there is no appeal data', async () => {
```
geek-soft, line 86: I guess that change in behaviour is fine as there was never requirement to display empty page and navigation system would not allow this to happen.  I even wonder whether we need test like that.

test/controllers/ConfirmationController.test.ts
```diff
@@ -22,11 +22,12 @@ describe('ConfirmationController', () => {
 
         const appeal = {
             penaltyIdentifier: {
-                companyNumber: '00345567',
+                companyNumber: 'NI000000',
```
geek-soft, line 25: I thought, wouldn't it be easier to change company number that was authenticated to `00345567`. Just a thought.

test/controllers/ConfirmationController.test.ts
```diff
@@ -22,11 +22,12 @@ describe('ConfirmationController', () => {
...
             },
         } as Appeal;
 
         const applicationData = {
+            appeal,
```
geek-soft, line 30: I see you included `appeal` property here. Will `appeal` still be present after submission? Isn't it moved to `submittedAppeal`? I am worrying that if `appeal` is not present it will break confirmation page.

test/controllers/OtherReasonController.test.ts
```diff
@@ -45,10 +51,17 @@ describe('OtherReasonController', () => {
...
+                    companyNumber: 'NI000000',
+                    penaltyReference: 'A00000001'
+                }
+            } as Appeal;
+
+            const app = createApp({ appeal });
 
             await request(app).post(OTHER_REASON_PAGE_URI)
-                .send({})
+                .send(appeal)
```
geek-soft, line 64: That seems wrong. Appeal isn't submitted when POST request is made on other reason page.

test/controllers/ConfirmationController.test.ts
n-r-b: you're right, the `appeal` is not available after submission, but the test is failing due to the CompanyAuthMiddleware expecting a company number on the `appeal` object. Thinking about it, the CompanyAuthMiddleware doesn't need to apply to this page since the appeal has already been submitted.. i'll clean this up

test/controllers/CheckYourAppealController.test.ts
n-r-b: the test was added when we first implemented auth and session middleware, i no longer feel it's relevant - i'll remove this

test/controllers/ConfirmationController.test.ts
geek-soft: If test is failing and appeal data is not going to be there that live app would misbehave as well. I think company auth middleware might not be needed for confirmation page cause its purpose is to stop people appealing on other people behalf without code and at that point appeal is already done.

n-r-b: after removing CompanyAuthMiddleware from the ConfirmationController, test passes. i've checked we're referencing the correct data

test/controllers/ConfirmationController.test.ts
n-r-b: yep! probably would have been a few less changes

PR #252, opened 2020-07-23
feat: Choose appeal reason (page and validation)

rhysbarrett:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1900

### Change description

- Implements the Choose Appeal Reason page, starting with a choice between "Illness" and "Other".
- Implements an environment variable-based Reason Whitelist, which acts as a feature toggle for this and future reason journeys.

**Note:** Rebuild will be necessary upon checkout due to new environment variable.

### Work checklist

### Merge instructions

```
```


.env.vagrant
```diff
@@ -34,6 +34,7 @@ OAUTH2_CLIENT_ID=
 OAUTH2_CLIENT_SECRET=
 
 COMPANY_AUTH_VERIFICATION_FEATURE_ENABLED=1
+ENABLED_APPEAL_REASONS=illness,other
```
geek-soft, line 37: This way feels confusing to me because removing 'other' from the list does not change much cause other reason screens will still be available, they might not be easily reachable from the reason list.

There is also no need to keep making list of reasons configurable, the feature flag is needed only to temporarily disable part of application under development in production until development is complete.

src/controllers/ChooseAppealReasonController.ts
```diff
@@ -0,0 +1,41 @@
...
+        return OTHER_REASON_PAGE_URI;
+    },
+    actions: (_: boolean) => {
+        return {
+            continue: 'action=continue'
+        };
+    }
+};
+
+@controller(CHOOSE_REASON_PAGE_URI, SessionMiddleware, AuthMiddleware, CompanyAuthMiddleware)
```
geek-soft, line 27: Path handled by the controller should be not available if feature is disabled. We do not want only hide options visually, we want completely disable incomplete feature.

src/controllers/validators/AppealReasonValidator.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export class AppealReasonValidator extends FormValidator {
+    constructor() {
+        super(schema);
+    }
+
+    public async validate(request: Request): Promise<ValidationResult> {
+        const result: ValidationResult = await super.validate(request);
+
+        request.body.reasons = createReasonsRadioGroup();
```
geek-soft, line 16: That seems bit like a random place to re-populate reasons list here. Validator should really be concerned with validation, nothing more.

In fact having to create separate validator which runs default validation logic and simply does totally unexpected thing is even weirder.

src/utils/FeatureChecker.ts
```diff
@@ -1,6 +1,12 @@
-import { getEnv } from 'app/utils/EnvironmentUtils';
+import { getEnv, getEnvOrDefault } from 'app/utils/EnvironmentUtils';
 import { Feature } from 'app/utils/Feature';
 
 export const isFeatureEnabled = (feature: Feature): boolean => {
     return getEnv(`${feature}_FEATURE`) === '1';
 };
+
+export const getEnabledAppealReasons = (): string[] => {
```
geek-soft, line 8: I do not think it is best place for that new function. Feature checker was meant to handle 0/1 feature flags while your approach is different and focused purely on appeal reasons.

I'd consider changing approach to have simple 0/1 feature flag for lines reason that will get removed in the future when feature is shipped.

src/views/choose-appeal-reason.njk
```diff
@@ -0,0 +1,59 @@
...
+
+{% block pageTitle %}
+  You must tell us the reason for the appeal
+{% endblock %}
+
+{% block backLink %}
+  {{ 
+    govukBackLink({
+      text: 'Back',
+      href: navigation.back.href
```
geek-soft, line 15: I do not think story defines that at this point and there are navigations stories in the backlog. I would not implement it in the similar stories.

test/controllers/ReviewPenaltyController.test.ts
```diff
@@ -72,7 +72,9 @@ describe('ReviewPenaltyController', () => {
 
     });
 
-    it('should go to other reasons disclaimer screen when continue is pressed', async () => {
+    it('should go to Choose Reason screen when continue is pressed', async () => {
```
geek-soft, line 75: That test should not be changed because a) story does not define navigation logic b) existing logic should be preserved as default and tested.

If conditional navigation was to be implemented it would require two tests, one for feature flag on and other for feature flag off.

test/models/ReasonsRadioGroup.test.ts
```diff
@@ -0,0 +1,18 @@
+import { expect } from 'chai';
+
+import { createReasonsRadioGroup } from 'app/models/components/ReasonsRadioGroup';
+
+describe('The Reasons Radio Group factory', () => {
+    it('should only return radio buttons for enabled reasons', () => {
```
geek-soft, line 6: One test feels bit light to be fair. What about env var? What about env var with more than one option?

Again this comment is only valid if you still to your approach. If you use 0/1 feature flag it might not be relevant anymore as most likely options would be hardcoded in NJK file.

test/controllers/ChooseAppealReasonController.test.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Appeal } from 'app/models/Appeal';
+import { ApplicationData } from 'app/models/ApplicationData';
+import { CHOOSE_REASON_PAGE_URI } from 'app/utils/Paths';
+
+describe('ChooseAppealReasonController', () => {
+
+    let env: NodeJS.ProcessEnv;
+
+    before( () => {
+        env = process.env;
```
geek-soft, line 14: That is becoming bit tedious as few tests already have this 1) capture envs 2) modify envs 3) restore envs approach. Can you propose something better? Maybe all future flags should be enabled by default for tests and only in rare instances value should be modified in the test?

src/models/components/ReasonsRadioGroup.ts
```diff
@@ -0,0 +1,15 @@
+import { getEnabledAppealReasons } from 'app/utils/FeatureChecker';
+
+export function createReasonsRadioGroup(): {}[] {
+    const enabledReasons = getEnabledAppealReasons();
+    const allReasons = [
+        { value: 'illness', text: 'Illness and health issues' },
+        { value: 'other', text: 'I\'m appealing for another reason' }
```
geek-soft, line 7: Use apostrophe instead of `\'`

src/controllers/validators/AppealReasonValidator.ts
rhysbarrett: For this, I'm following the pattern established in `src/controllers/MultiplePenaltiesValidator`. The code smells to me as well, but I tried many different ways to populate the reasons list on POST and none of them worked without breaking tests elsewhere in the system.

src/controllers/ChooseAppealReasonController.ts
rhysbarrett: Okay, I'll look into disabling the route. Would a simple redirect suffice or do I need to include logic to completely disable to controller when the reason isn't enabled?

.env.vagrant
rhysbarrett: This is the most compact solution I could think of for allowing the Choose Reasons page to dynamically enable/disable reasons. While this isn't high-priority for the current journey (because disabling Illnesses should simply disable the entire Choose Reason page), it makes it far easier to enable/disable upcoming reason journeys during their development.

I agree that having "other" in there is unnecessary though. I can include some additional logic so that "other" is included by default.

src/utils/FeatureChecker.ts
rhysbarrett: Okay, I'll rework solution and refactor accordingly. This will only work however if the assumption that only one reason journey will ever be in development at any one time holds true.

src/views/choose-appeal-reason.njk
rhysbarrett: I can't figure out how to implement validation without extending the SafeNavigationBaseController, which in turn requires the navigation object. My thought process is that I'd rather do a small amount of future work if it means I can avoid a significantly greater amount of work of figuring out how to how to do something in a way that will be completely reworked anyway as soon as the navigation tickets come into play.

test/controllers/ReviewPenaltyController.test.ts
rhysbarrett: As mentioned in previous comment, I can't figure out how to complete the story without extending the SafeNavigationBaseController. With this in mind, I can't access the page without first hitting Continue on a previous page, which according to the Prototype is this one, hence the navigation update. 

I will however get the second condition tested.

test/models/ReasonsRadioGroup.test.ts
rhysbarrett: I'll look into additional test cases, unless we decide to change the feature flag approach.

test/controllers/ChooseAppealReasonController.test.ts
rhysbarrett: How do I modify the default values for test environment variables such that they aren't dependent on the runtime environment variables? My understanding is that the test environment should be consistent and independent from runtime assumptions.

src/models/components/ReasonsRadioGroup.ts
rhysbarrett: I'm unsure what you mean here. An apostrophe would terminate the string prematurely without the escape character. Do you mean I should swap the enclosing apostrophes for quotation marks?

geek-soft: No, I mean proper apostrophe character (`’`) that you get from Option + Shift + ] combination.

.env.vagrant
rhysbarrett: Resolved: Switched to binary toggle

src/controllers/validators/AppealReasonValidator.ts
rhysbarrett: Resolved: Removed file completely, template manipulation no longer required due to binary feature toggle.

src/views/choose-appeal-reason.njk
rhysbarrett: Resolved: Controller now extends BaseAsyncController so navigation object is no longer required.

test/controllers/ReviewPenaltyController.test.ts
rhysbarrett: Resolved: Navigation is no longer affected by this feature so the test has been unaltered.

test/models/ReasonsRadioGroup.test.ts
rhysbarrett: Resolved: Component factory has been removed as template is now static.

src/utils/FeatureChecker.ts
rhysbarrett: Resolved: Switched to binary feature flag as per @n-r-b 's changes.

src/controllers/ChooseAppealReasonController.ts
rhysbarrett: Resolved: Route disabling implemented by @n-r-b 's feature toggle middleware.

test/controllers/ChooseAppealReasonController.test.ts
rhysbarrett: Resolved: Environment manipulation is no longer required due to change in feature toggle approach.

src/models/components/ReasonsRadioGroup.ts
rhysbarrett: Oh, that's a backtick, but okay. This has been refactored into the template now anyway. Marking as resolved.

src/controllers/ChooseAppealReasonController.ts
```diff
@@ -0,0 +1,35 @@
...
+    public async renderView(): Promise<void> {
+        return this.render(template);
+    }
+
+    @httpPost('')
+    public async continue(): Promise<void> {
+        const request = this.httpContext.request;
+        const validationResult: ValidationResult = await new FormValidator(schema).validate(request);
+
+        if (validationResult.errors.length > 0) {
```
geek-soft, line 28: No persistence in session. I'd use `BaseController` to get these for free.

rhysbarrett: After a _long_ discussion, agreed and changed. Marking as resolved.

src/controllers/ChooseAppealReasonController.ts
```diff
@@ -0,0 +1,34 @@
...
+        return '';
+    },
+    actions: (_: boolean) => {
+        return {
+            continue:'action=continue'
+        };
+    }
+};
+
+@controller(CHOOSE_REASON_PAGE_URI, IllnessReasonFeatureMiddleware, SessionMiddleware, AuthMiddleware)
```
geek-soft, line 28: Shall we add company auth middleware to not forget? I do not think there is story for that.

test/utils/FeatureChecker.test.ts
```diff
@@ -1,4 +1,4 @@
-import { expect } from 'chai';
+import { assert } from 'chai';
```
geek-soft, line 1: We tend to user chai/expect. I do not quite see point for making that change.

src/views/choose-appeal-reason.njk
```diff
@@ -0,0 +1,62 @@
...
+
+{% from 'govuk/components/back-link/macro.njk'     import govukBackLink %}
+{% from 'govuk/components/button/macro.njk' import govukButton %}
+{% from 'govuk/components/radios/macro.njk' import govukRadios %}
+
+{% block pageTitle %}
+  You must tell us the reason for the appeal
+{% endblock %}
+
+{% block backLink %}
```
geek-soft, line 11: In the future you do not need to add back link if navigation is out side of scope. It just adds confusion to code review.

src/views/choose-appeal-reason.njk
```diff
@@ -0,0 +1,62 @@
...
+                                        classes: 'govuk-fieldset__legend--xl'
+                                    }
+                                },
+                                hint: {
+                                    text: 'You can add more reasons later'
+                                },
+                                errorMessage: validationResult.getErrorForField('reason') if validationResult,
+                                items: [
+                                    { value: 'illness', text: 'Illness and health issues' },
+                                    { value: 'other', text: "I'm appealing for another reason" }
```
geek-soft, line 44: Wrap text in single quote and use proper apostrophe (Shift + Option + ]).

src/models/Illness.ts
```diff
@@ -0,0 +1,14 @@
+
+import { IllPerson } from './fields/IllPerson';
+
+import { Attachment } from 'app/models/Attachment';
+
+export interface Illness {
```
geek-soft, line 6: These data models are not part of the ticket, remove for better traceability.

src/controllers/ChooseAppealReasonController.ts
rhysbarrett: Adding the CompanyAuth middleware would tie the page into the journey, as attempting to access the page without authenticating first would cause a redirect. To my understanding, for this ticket we should be able to access the page simply by entering the URL. If this isn't the case then I'm happy to make the change.

test/utils/FeatureChecker.test.ts
rhysbarrett: I made this change originally when I added that `getEnabledReasons` function, because I noticed all of the TSLint issues with the existing tests. I figured since both test methods are from the same library, the change was a more elegant solution than simply disabling the linter.

src/views/choose-appeal-reason.njk
rhysbarrett: No problem, I was attempting to replicate the page on the prototype. Would you like me to remove it in this instance or just bear in mind for future stories?

src/views/choose-appeal-reason.njk
rhysbarrett: Understood and addressed. I didn't realise that was an option (thought it was a different way of entering a backtick when you mentioned it last time) so thanks for pointing that out!

src/models/Illness.ts
rhysbarrett: Sure.

src/controllers/ChooseAppealReasonController.ts
geek-soft: You are right although it is same with being logged in. It does not matter much as long as Richard knows that he needs to authenticate company before entering that screen. If you do not add it make sure it is added somewhere so we do not forget to add it in the future.

rhysbarrett: If I don't add it now, I can add a comment to the Nav ticket to remind whoever picks that up to add it.

src/views/choose-appeal-reason.njk
geek-soft: Bear in mind. Because how we structured stories every new joiner adds that link to the first story ;)

test/utils/FeatureChecker.test.ts
geek-soft: I will use consistency card this time. If you do not like suppressing linter warning you could use `be.equal(true)`.

src/controllers/ChooseAppealReasonController.ts
rhysbarrett: Okie dokie. I've added a note to the relevant Jira ticket.

src/views/choose-appeal-reason.njk
rhysbarrett: Okie dokie!

test/utils/FeatureChecker.test.ts
rhysbarrett: Sure, I'll revert the change then.

PR #253, opened 2020-07-27
Feature/lfa 1439 illness start date

n-r-b:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1439

### Change description
Implemented the following:-
- New illness start date template:
Screenshot 2020-07-27 at 09 38 47

- Required fields validation:
Screenshot 2020-07-27 at 09 39 41

- Invalid (non-existing) date validation:
Screenshot 2020-07-27 at 09 41 24

- Invalid (in future) date validation:
Screenshot 2020-07-27 at 09 42 17

- Illness reason feature flag middleware 

### Work checklist

- [x ] Tests added where applicable
- [x ] UI changes look good on mobile
- [x ] UI changes meet accessibility criteria

### Merge instructions

```
```


src/models/fields/IllnessStartDate.schema.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const schema = Joi.object({
+    startDay: Joi.string()
+        .required()
+        .pattern(dayMonthRegex)
+        .messages({
+            'any.required': illnessStartDayErrorMessage,
+            'string.base': illnessStartDayErrorMessage,
+            'string.empty': illnessStartDayErrorMessage,
+            'string.pattern.base': illnessStartDayErrorMessage
```
akaJSaunders, line 20: doesn't your pattern cover the `string.base`, `string.empty` and `any.required`?

src/views/illness/illness-start-date.njk
```diff
@@ -0,0 +1,111 @@
+{% extends 'layout.njk' %}
+
+{% from 'govuk/components/back-link/macro.njk' import govukBackLink %}
+{% from 'govuk/components/error-summary/macro.njk' import govukErrorSummary %}
+{% from 'govuk/components/button/macro.njk' import govukButton %}
+{% from 'govuk/components/date-input/macro.njk' import govukDateInput %}
+
+{% block pageTitle %}
+   Appeal due to ill health: Tell us when the illness started
```
akaJSaunders, line 9: Is this page title stated in the ticket? Tell has a capital T even though its not starting a sentence, is that how its meant to be?

src/views/illness/illness-start-date.njk
```diff
@@ -0,0 +1,111 @@
...
+        {% if isStartDateMonthErrorFlag %}
+            {% set classes_month = classes_month + ' govuk-input--error' %}
+          {% endif %}
+        {% if isStartDateYearErrorFlag %}
+          {% set classes_year = classes_year + ' govuk-input--error' %}
+        {% endif %}
+
+        {% if validationResult and validationResult.errors.length > 0 %}
+          {% set errorMessages %}
+            <ul style='padding: 0;list-style-type: none;'>
```
akaJSaunders, line 40: I think we have been using double quotes for html code

src/views/illness/illness-start-date.njk
```diff
@@ -0,0 +1,111 @@
...
+          {% endset %}
+
+          {% if errorMessages %}
+            {% set illnessStartErrorMessage = {
+              html: errorMessages
+             } %}
+          {% else %}
+            {% set illnessStartErrorMessage = false %}
+          {% endif %}
+        {% else %}
```
akaJSaunders, line 56: what action does this `{% else %}` perform?

src/views/illness/illness-start-date.njk
```diff
@@ -0,0 +1,111 @@
...
+        {% endif %}
+
+        {{ govukDateInput({
```
akaJSaunders, line 59: Could you format this similar to how you did it in lines 17-22?

src/views/illness/illness-start-date.njk
```diff
@@ -0,0 +1,111 @@
...
+            {
+              classes: classes_year,
+              id: 'start-year',
+              label: 'Year',
+              name: 'startYear',
+              value: startYear,
+              autocomplete: 'off'
+            }
+          ]
+        }) }}
```
akaJSaunders, line 98: Could you format this similar to how you did it in lines 17-22?

.env.vagrant
```diff
@@ -17,6 +17,7 @@ SUPPORTED_MIME_TYPES='application/pdf,image/jpeg,image/png,application/vnd.openx
 MAX_NUMBER_OF_FILES=10
 
 FILE_TRANSFER_FEATURE=1
+ILLNESS_REASON_FEATURE=1
```
akaJSaunders, line 20: Has variable also been added to the chs-config and docker development repos?

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,75 @@
...
+import { IllnessReasonFeatureMiddleware } from 'app/middleware/IllnessReasonFeatureMiddleware';
+import { schema } from 'app/models/fields/IllnessStartDate.schema';
+import { ILLNESS_START_DATE_PAGE_URI } from 'app/utils/Paths';
+import { ValidationResult } from 'app/utils/validation/ValidationResult';
+
+const template = 'illness/illness-start-date';
+const ILLNESS_START_DAY_FIELD = 'startDay';
+const ILLNESS_START_MONTH_FIELD = 'startMonth';
+const ILLNESS_START_YEAR_FIELD = 'startYear';
+const ILLNESS_FULL_DATE = 'startDate';
```
akaJSaunders, line 16: why are these capitalised?

src/models/fields/IllnessStartDate.schema.ts
```diff
@@ -0,0 +1,50 @@
+import * as Joi from '@hapi/joi';
+
+const illnessStartDayErrorMessage: string = 'You must enter a day';
+const illnessStartMonthErrorMessage: string = 'You must enter a month';
+const illnessStartYearErrorMessage: string = 'You must enter a year';
+const invalidDate: string = 'Enter a real date';
+const invalidDateFuture: string = 'Start date must be today or in the past';
```
akaJSaunders, line 7: you added to the variables of some of these constants "ErrorMessage" at the end of the name, but not with invalidDate and invalidDateFuture

```suggestion
const illnessStartDayErrorMessage: string = 'You must enter a day';
const illnessStartMonthErrorMessage: string = 'You must enter a month';
const illnessStartYearErrorMessage: string = 'You must enter a year';
const invalidDateErrorMessage: string = 'Enter a real date';
const invalidDateFutureErrorMessage: string = 'Start date must be today or in the past';
```

src/models/fields/IllnessStartDate.schema.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const dayMonthRegex: RegExp = /^[0-9]{1,2}$/i;
+const yearRegex: RegExp = /^[0-9]{2,4}$/i;
```
akaJSaunders, line 10: do we really want to allow 2 characters for years?

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,75 @@
...
+                        break;
+                }
+            });
+            return this.renderWithStatus(UNPROCESSABLE_ENTITY)(
+                template, {
+                    validationResult,
+                    isStartDateDayErrorFlag: startDateDayErrorFlag,
+                    isStartDateMonthErrorFlag: startDateMonthErrorFlag,
+                    isStartDateYearErrorFlag: startDateYearErrorFlag,
+                });
```
akaJSaunders, line 70: In your screenshots, the errors seem to be displayed correctly, but the invalid data that the user previously would have put in, is erased. This behavior is different to how we implemented the penalty reference page, example bellow:

Screenshot 2020-07-27 at 13 28 46

.env.vagrant
akaJSaunders, line 20: Where does the environment variable `ILLNESS_REASON_FEATURE` get read?

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,156 @@
...
+const invalidDateErrorMessage: string = 'Enter a real date';
+const invalidDateFutureErrorMessage: string = 'Start date must be today or in the past';
+
+describe('IllnessStartDateController', () => {
+
+    describe('GET request', () => {
+
+        before(() => {
+            process.env.ILLNESS_REASON_FEATURE = '1';
+        });
```
akaJSaunders, line 28: `before` in this case only is applied to two tests, but one of them overides it by seting the feature to 0. So this is only really useful for one test, id move this inside that test

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,156 @@
...
+                    expect(response.status).to.be.equal(MOVED_TEMPORARILY);
+                    expect(response.header.location).to.include(ENTRY_PAGE_URI);
+                });
+        });
+    });
+    describe('POST request', () => {
+
+        before(() => {
+            process.env.ILLNESS_REASON_FEATURE = '1';
+        });
```
akaJSaunders, line 54: does this need to be set before every test? or could it just be:

```suggestion
        process.env.ILLNESS_REASON_FEATURE = '1';
```

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,156 @@
...
+        it('should return 200 when posting a valid date', async () => {
+            const app = createApp();
+            await request(app).post(ILLNESS_START_DATE_PAGE_URI)
+                .send({startDay: '01', startMonth: '01', startYear: '2020'})
+                .expect(response => {
+                    expect(response.status).to.be.equal(OK);
+                    expect(response.text).to.contain(pageHeading);
+                });
+        });
+        it('should return 422 response with rendered error messages when empty start date was submitted', async () => {
```
akaJSaunders, line 65: could we have an empty line in between every test?

src/models/fields/IllnessStartDate.schema.ts
akaJSaunders: (if we are meant to allow for years with only 2 characters, could we have a test for that please)

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,156 @@
...
+                        expect(response.text).to.include(pageHeading)
+                            .and.to.include(errorSummaryHeading)
+                            .and.to.include(invalidDateErrorMessage);
+                    });
+            });
+        it('should return 422 response with rendered error message invalid start date (in future) was submitted',
+            async () => {
+                const app = createApp();
+                await request(app).post(ILLNESS_START_DATE_PAGE_URI)
+                    .send({startDay: '01', startMonth: '01', startYear: '2029'})
```
akaJSaunders, line 135: If this service survives 2029, would that mean this test would fail? Don't think we need to address that, but I thought it was a fun thought :P

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+            process.env.ILLNESS_REASON_FEATURE = '0';
+
+            const app = createApp();
+            await request(app).post(ILLNESS_START_DATE_PAGE_URI)
+                .send({})
+                .expect(response => {
+                    expect(response.status).to.be.equal(MOVED_TEMPORARILY);
+                    expect(response.header.location).to.include(ENTRY_PAGE_URI);
+                });
```
akaJSaunders, line 153: isnt this already tested?

test/middleware/IllnessFeatureMiddleware.test.ts
```diff
@@ -0,0 +1,86 @@
...
+import { ENTRY_PAGE_URI, ILLNESS_START_DATE_PAGE_URI } from 'app/utils/Paths';
+
+import { createApp } from 'test/ApplicationFactory';
+
+let initialIllnessReasonFeatureFlag: string | undefined;
+
+describe('Illness Reason Feature Toggle Middleware', () => {
+
+    before(() => {
+        initialIllnessReasonFeatureFlag = process.env.ILLNESS_REASON_FEATURE;
```
akaJSaunders, line 20: Are you reading from the .env here?

test/middleware/IllnessFeatureMiddleware.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    after(() => {
+        process.env.ILLNESS_REASON_FEATURE = initialIllnessReasonFeatureFlag;
+    });
+
+    describe('Feature switched on', () => {
+
+        it('should call next() if feature flag is on', () => {
+
+            process.env.ILLNESS_REASON_FEATURE = '1';
```
akaJSaunders, line 31: does `@before` get triggered before the test starts? if you set the process.env here, whats the points of the:


    before(() => {
        initialIllnessReasonFeatureFlag = process.env.ILLNESS_REASON_FEATURE;
    });

    after(() => {
        process.env.ILLNESS_REASON_FEATURE = initialIllnessReasonFeatureFlag;
    });

test/middleware/IllnessFeatureMiddleware.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+            const mockRequest = Substitute.for<Request>();
+            const mockResponse = Substitute.for<Response>();
+            const mockNext = Substitute.for<NextFunction>();
+
+            new IllnessReasonFeatureMiddleware().handler(mockRequest, mockResponse, mockNext);
+
+            // @ts-ignore
```
akaJSaunders, line 39: what are we ignoring here?

test/middleware/IllnessFeatureMiddleware.test.ts
```diff
@@ -0,0 +1,86 @@
...
+            process.env.ILLNESS_REASON_FEATURE = '0';
+
+            const mockRequest = Substitute.for<Request>();
+            const mockResponse = Substitute.for<Response>();
+            const mockNext = Substitute.for<NextFunction>();
+
+            new IllnessReasonFeatureMiddleware().handler(mockRequest, mockResponse, mockNext);
+
+            // @ts-ignore
+            mockResponse.received().redirect(Arg.any());
```
akaJSaunders, line 69: is it maybe a good idea to also test that the user is being redirected to the right location?

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,75 @@
...
+import { ValidationResult } from 'app/utils/validation/ValidationResult';
+
+const template = 'illness/illness-start-date';
+const ILLNESS_START_DAY_FIELD = 'startDay';
+const ILLNESS_START_MONTH_FIELD = 'startMonth';
+const ILLNESS_START_YEAR_FIELD = 'startYear';
+const ILLNESS_FULL_DATE = 'startDate';
+
+@controller(ILLNESS_START_DATE_PAGE_URI, SessionMiddleware, IllnessReasonFeatureMiddleware)
+export class IllnessStartDateController extends BaseAsyncHttpController {
```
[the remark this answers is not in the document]
n-r-b: the `BaseHttpController` or the `SafeNavigationBaseController`? I'm using the `BaseAsyncHttpController` which extends the `BaseHttpController` and provides us with reusable render methods. I'm not using the `SafeNavigationBaseController` as navigation is not part of this ticket

src/views/illness/illness-start-date.njk
n-r-b: I've taken this from the prototype

src/views/illness/illness-start-date.njk
n-r-b: that is not needed! i'll remove it

.env.vagrant
n-r-b: not yet ;) but thanks for the reminder

src/controllers/IllnessStartDateController.ts
n-r-b: i'm used to constants being capitalised especially when their value does not change. But i'm happy to make these lowercase to be consistent with the project

.env.vagrant
n-r-b: The `IllnessReasonFeatureMiddleware` uses the `isFeatureEnabled` const in the `FeatureChecker` class to get the env variable

n-r-b: it appends `_FEATURE` to property

src/models/fields/IllnessStartDate.schema.ts
n-r-b: i'll check this

src/middleware/IllnessReasonFeatureMiddleware.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Feature } from 'app/utils/Feature';
+import { isFeatureEnabled } from 'app/utils/FeatureChecker';
+import { ENTRY_PAGE_URI } from 'app/utils/Paths';
+
+@provide(IllnessReasonFeatureMiddleware)
+export class IllnessReasonFeatureMiddleware extends BaseMiddleware {
+
+    public handler(req: Request, res: Response, next: NextFunction): void {
+
+        if (isFeatureEnabled(Feature.ILLNESS_REASON)) {
```
geek-soft, line 15: Feels like this logic is duplicated in at least one more place. Wouldn't it be better top have one configurable middleware that is tested once and just configured in places it is needed?

src/controllers/IllnessStartDateController.ts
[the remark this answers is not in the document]
geek-soft: Why not to use BaseController as in most places?

.env.vagrant
geek-soft, line 20: We are bing bit inconsistent in making toggles.

We have two formats:
1. `COMPANY_AUTH_VERIFICATION_FEATURE_ENABLED`
2. `ILLNESS_REASON_FEATURE`

I prefer the first one due to being more descriptive but we should strive to keep one format only. We should also remove file upload toggle.

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,75 @@
...
+                        if (startDateDayErrorFlag || startDateMonthErrorFlag || startDateYearErrorFlag) {
+                            validationResult.errors.splice(validationResult.errors.indexOf(err), 1);
+                        }
+                        startDateDayErrorFlag = true;
+                        startDateMonthErrorFlag = true;
+                        startDateYearErrorFlag = true;
+                        break;
+                }
+            });
+            return this.renderWithStatus(UNPROCESSABLE_ENTITY)(
```
geek-soft, line 64: Where is session handling logic? Data should be persistent in session, shouldn't it?

src/models/fields/IllnessStartDate.schema.ts
```diff
@@ -0,0 +1,50 @@
+import * as Joi from '@hapi/joi';
+
+const illnessStartDayErrorMessage: string = 'You must enter a day';
+const illnessStartMonthErrorMessage: string = 'You must enter a month';
+const illnessStartYearErrorMessage: string = 'You must enter a year';
+const invalidDate: string = 'Enter a real date';
+const invalidDateFuture: string = 'Start date must be today or in the past';
+
+const dayMonthRegex: RegExp = /^[0-9]{1,2}$/i;
```
geek-soft, line 9: `i` modifier making regex case insensitive does not make much sense here since only numbers are allowed, does it?

test/controllers/IllnessStartDateController.test.ts
n-r-b: i've added a before() and after() to retain the feature state at class level, and have then added a beforeEach() to enable to feature for the tests

src/models/fields/IllnessStartDate.schema.ts
```diff
@@ -0,0 +1,50 @@
...
+const yearRegex: RegExp = /^[0-9]{2,4}$/i;
+
+export const schema = Joi.object({
```
geek-soft, line 12: That seems like reusable date schema, so it could be more reusable if done the way src/models/fields/YesNo.schema.ts was implemented.

test/middleware/IllnessFeatureMiddleware.test.ts
n-r-b: removed

src/views/illness/illness-start-date.njk
```diff
@@ -0,0 +1,111 @@
...
+            titleText: 'There is a problem',
+            errorList: validationResult.errors
+          }) if validationResult and validationResult.errors.length > 0
+        }}
+
+        {% set classes_day = 'govuk-input--width-2' %}
+        {% set classes_month = 'govuk-input--width-2' %}
+        {% set classes_year = 'govuk-input--width-4' %}
+
+        {% if isStartDateDayErrorFlag %}
```
geek-soft, line 28: I feel this and controller logic are super complex. Would't that be easier to just do `validationResult.getErrorForField('startDay') or validationResult.getErrorForField('startDate')` in template?

src/views/illness/illness-start-date.njk
geek-soft: Formatting is inconsistent, it should be more:

```
{{
  govukDateInput({
...
```

src/models/fields/IllnessStartDate.schema.ts
n-r-b: i'm not 100% sure about this.. the example does suggest 4 characters - i'll follow the example as that's all we have

src/views/illness/illness-start-date.njk
n-r-b: i've removed the flags and refactored logic in controller/template

src/controllers/IllnessStartDateController.ts
n-r-b: this should have been resolved now

src/controllers/IllnessStartDateController.ts
n-r-b: this has now been added

src/models/fields/IllnessStartDate.schema.ts
n-r-b: tried removing the above and the tests then fail, as an alternative error message is returned

test/controllers/IllnessStartDateController.test.ts
n-r-b: good point - updated test

test/middleware/IllnessFeatureMiddleware.test.ts
n-r-b: The before() and after() tests retain the env property value. 
Each time we set `process.env.ILLNESS_REASON_FEATURE=0` we override this value for the application, so having the after() test allows us to reset the value to what it was before

test/middleware/IllnessFeatureMiddleware.test.ts
n-r-b: yes, see comment below

test/controllers/IllnessStartDateController.test.ts
n-r-b: i've removed the these duplicate controller integration tests from the middleware test

test/middleware/IllnessFeatureMiddleware.test.ts
n-r-b: done

src/middleware/IllnessReasonFeatureMiddleware.ts
n-r-b: Good idea - i've created a configurable middleware, see `FeatureToggleMiddleware.ts`

.env.vagrant
n-r-b: Yes, i agree with having a consistent format across the app. I've updated the feature toggles to the 1st format. I'll remove the file upload toggle in another PR

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,208 @@
...
+                    companyNumber: 'NI000000',
+                    penaltyReference: 'A00000001'
+                },
+                reasons: {
+                    other: {
+                        title: 'I have reasons',
+                        description: 'they are legit'
+                    },
+                    illness: {
+                        startDate: '2020-01-01T00:00:00.000Z'
```
rhysbarrett, line 65: Are we keeping the time component? If so I'll need to update my DateFilter.

.env.vagrant
```diff
@@ -34,7 +31,9 @@ OAUTH2_TOKEN_URI=http://web.chs-dev.internal:4000/oauth2/token
 OAUTH2_CLIENT_ID=
 OAUTH2_CLIENT_SECRET=
 
+FILE_TRANSFER_FEATURE_ENABLED=1
```
geek-soft, line 34: Do not forget to remove it when refactor is merged in. I suspect there might be code conflict around this place.

src/ContainerFactory.ts
```diff
@@ -101,6 +102,8 @@ export function createContainer(): Container {
             sessionConfig,
             companyAuthFeatureEnabled));
 
+    container.bind(FeatureToggleMiddleware).toConstantValue(FeatureToggleMiddleware);
```
geek-soft, line 105: That does not make much sense and I believe it is not needed cause you are not injecting instances of `FeatureToggleMiddleware` due to required argument.

By the way I'd love to have separate PR that refactors feature toggle after one for file upload was removed. That would keep PR small and focus to me. I am not sure through how much work it would be to split it now, if too much then not a point.

.env.vagrant
```diff
@@ -34,6 +32,7 @@ OAUTH2_CLIENT_ID=
 OAUTH2_CLIENT_SECRET=
 
 COMPANY_AUTH_VERIFICATION_FEATURE_ENABLED=1
+ILLNESS_REASON_FEATURE_ENABLED=1
```
geek-soft, line 35: I love the fact you made it consistent, although remember changing it in two more config repositories.

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,64 @@
...
+    previous(): string {
+        return ILLNESS_START_DATE_PAGE_URI;
+    },
+    next(): string {
+        return ILLNESS_START_DATE_PAGE_URI;
+    }
+};
+
+const applyPadding = (dayMonth: string): string => {
+    return dayMonth.length < 2 ? dayMonth.padStart(2, '0') : dayMonth;
```
geek-soft, line 28: That condition is not needed as per screenshot below:

image

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+@controller(ILLNESS_START_DATE_PAGE_URI, SessionMiddleware, AuthMiddleware,
```
geek-soft, line 31: Feels like FeatureToggleMiddleware should be first middleware that is executed. No point checking session / auth if feature is disabled.

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,64 @@
...
+    FeatureToggleMiddleware(Feature.ILLNESS_REASON))
+export class IllnessStartDateController extends BaseController<Illness> {
```
geek-soft, line 33: Using `Illness` interface here is not quite correct cause `Illness` class has many more properties than what is available via `request.body` (form data structure).

In other PR it was implemented as:

image

In that instance internal interface describes data structure that HTML form created.

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    constructor() {
+        super(template, navigation, new StartDateValidator());
+    }
+
+    protected prepareViewModelFromAppeal(appeal: Appeal): any {
+        const illness: Illness | undefined = appeal.reasons?.illness || undefined;
```
geek-soft, line 40: A ` || undefined` construct is not needed IMO cause it will be undefined anyway if undefined :D

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,64 @@
...
+        const day: string = applyPadding(startDate.getDate().toString());
+        const month: string = applyPadding((startDate.getMonth() + 1).toString());
+        const year: string = startDate.getFullYear().toString();
+
+        return {day, month, year};
+    }
+
+    protected prepareSessionModelPriorSave(appeal: Appeal, value: Illness): Appeal {
+        if (appeal.reasons?.illness != null) {
+            appeal.reasons.illness = value;
```
geek-soft, line 54: This is not quite correct. Due to the fact `appeal.reasons?.illness` has more properties than just start date, in case illness object already exists this would override properties captured on other screens such as illness continuation or illness end date. It should be more:

```
appeal.reasons.illness.illnessStart = illnessStart;
```

src/models/fields/Date.schema.ts
```diff
@@ -0,0 +1,50 @@
...
+    year: Joi.string()
+        .required()
+        .pattern(yearRegex)
+        .messages({
+            'any.required': invalidYearErrorMessage,
+            'string.base': invalidYearErrorMessage,
+            'string.empty': invalidYearErrorMessage,
+            'string.pattern.base': invalidYearErrorMessage
+        }),
+    illnessStart: Joi.date()
```
geek-soft, line 40: That should be more generic to keep schema reusable for other screens like illness end date. It could be called `date`.

src/controllers/validators/StartDateValidator.ts
```diff
@@ -0,0 +1,38 @@
+import { Request } from 'express';
+import { FormValidator } from './FormValidator';
+
+import { schema } from 'app/models/fields/Date.schema';
+import { ValidationError } from 'app/utils/validation/ValidationError';
+import { ValidationResult } from 'app/utils/validation/ValidationResult';
+
+export class StartDateValidator extends FormValidator {
```
geek-soft, line 8: Same here, class could be named in more generic way to make it reusable for other date screens.

src/views/illness/illness-start-date.njk
```diff
@@ -0,0 +1,119 @@
...
+            <ul style="padding: 0;list-style-type: none;">
+              {% for error in validationResult.errors %}
+                <li>
+                  {{ error.text }}
+                </li>
+              {% endfor %}
+            </ul>
+          {% endset %}
+
+          {% if errorMessages %}
```
geek-soft, line 55: That 'if' statement is not clear to me. When I read it I think it is always set due to `set` clause above. If so there is no point having 'if' statement.

test/models/IllnessStartDateSchema.test.ts
```diff
@@ -0,0 +1,157 @@
...
+import { assertValidationErrors } from 'test/models/ValidationAssertions';
+
+const validator = new SchemaValidator(schema);
+
+const dayField: string = 'day';
+const monthField: string = 'month';
+const yearField: string = 'year';
+const startDateField: string = 'illnessStart';
+
+describe('IllnessStartDate schema', () => {
```
geek-soft, line 14: Again, I suggest rename if we plan to make date schema reusable.

test/models/IllnessStartDateSchema.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    it('should reject empty object', () => {
+        const validationResult = validator.validate({});
+        assertValidationErrors(validationResult, [
+            new ValidationError(dayField, 'You must enter a day'),
+            new ValidationError(monthField, 'You must enter a month'),
+            new ValidationError(yearField, 'You must enter a year'),
+            new ValidationError(startDateField, 'Enter a real date')
```
geek-soft, line 22: This is bit confusing of course and you work around it in the validator but maybe in the future we could try conditional schema so that either three ingredient fields are returned or one for full date.

test/models/IllnessStartDateSchema.test.ts
```diff
@@ -0,0 +1,157 @@
...
+            illnessStart: '0'
+        });
+        assertValidationErrors(validationResult, [
+            new ValidationError(dayField, 'You must enter a day'),
+            new ValidationError(monthField, 'You must enter a month'),
+            new ValidationError(yearField, 'You must enter a year'),
+            new ValidationError(startDateField, 'Enter a real date')
+        ]);
+    });
+
```
geek-soft, line 100: Unnecessary blank line here and in one place below. In turn there is no blank line before last test case.

test/models/IllnessStartDateSchema.test.ts
```diff
@@ -0,0 +1,157 @@
...
+            year: '2020',
+            illnessStart: '2020-01-33'
+        });
+        assertValidationErrors(validationResult, [
+            new ValidationError(startDateField, 'Enter a real date')
+        ]);
+    });
+
+
+    it('should reject invalid (format) date', () => {
```
geek-soft, line 115: What is wrong with that date? Seems legit to me.

src/models/fields/Date.schema.ts
```diff
@@ -0,0 +1,50 @@
...
+        .required()
+        .iso()
+        .max('now')
```
geek-soft, line 43: Maybe that could be configurable? To make it more reusable.

test/models/IllnessStartDateSchema.test.ts
geek-soft: Alright, it is not about format, it is because date is in the past. Test name could be more descriptive.

geek-soft: Is that because type is wrong? If so I got confused cause other examples that were invalid used string date as well.

test/models/IllnessStartDateSchema.test.ts
```diff
@@ -0,0 +1,157 @@
...
+            illnessStart: new Date('2020-01-01')
+        });
+        assertValidationErrors(validationResult, []);
+    });
+    it('should accept single digit date values', () => {
+        const validationResult = validator.validate({
+            day: '1',
+            month: '1',
+            year: '2020',
+            illnessStart: new Date('2020-01-01')
```
geek-soft, line 153: This date validation feels so artificial due to the fact both date ingredients and full date is validated separately while values should be in sync. Crazy. As a result we have a test that tests single digits values but ISO date is correct ISO date. I understand validator will prepare data for validation but looking at that tests gives me mixed feelings :(

src/models/fields/Date.schema.ts
n-r-b: i'm happy to make this generic, but then the `.max('now')` validation may not apply to an end date for example. But i'm sure there's a way to make this configurable when it comes to this

test/models/IllnessStartDateSchema.test.ts
n-r-b: the type should be a valid ISO 8601 date format - i'll rename the test

src/models/fields/Date.schema.ts
n-r-b: yes that would be good, i'll see if i can do this

src/controllers/IllnessStartDateController.ts
n-r-b: done

src/views/illness/illness-start-date.njk
n-r-b: good point, i've removed the if statement now

src/controllers/IllnessStartDateController.ts
n-r-b: removed the padding - just thought it would look nicer when retrieving date from the session

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,67 @@
...
+        }
+        const startDate: Date = new Date(illness.illnessStart);
+        const day: string = startDate.getDate().toString();
+        const month: string = (startDate.getMonth() + 1).toString();
+        const year: string = startDate.getFullYear().toString();
+
+        return {day, month, year};
+    }
+
+    protected prepareSessionModelPriorSave(appeal: Appeal, value: any): Appeal {
```
geek-soft, line 52: Change `any` to `FormBody`.

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,67 @@
...
+    previous(): string {
+        return ILLNESS_START_DATE_PAGE_URI;
+    },
+    next(): string {
+        return ILLNESS_START_DATE_PAGE_URI;
+    }
+};
+
+interface FormBody {
+    illnessStart: string;
```
geek-soft, line 28: This does not match actual values in `request.body`. It should have `date: Date` there.

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,67 @@
...
+        const month: string = (startDate.getMonth() + 1).toString();
+        const year: string = startDate.getFullYear().toString();
+
+        return {day, month, year};
+    }
+
+    protected prepareSessionModelPriorSave(appeal: Appeal, value: any): Appeal {
+        const illness: Illness | undefined = appeal.reasons.illness;
+        if (illness != null) {
+            illness.illnessStart = value.date;
```
geek-soft, line 55: This is wired. In the `Illness` interface `illnessStart` is defined as string while `value.date` is defined as Date.

src/models/fields/Date.schema.ts
geek-soft: Let's do it when we approach illness end date.

test/models/chunks/DateSchema.test.ts
```diff
@@ -0,0 +1,157 @@
...
+            year: '2020',
+            date: '2020-01-33'
+        });
+        assertValidationErrors(validationResult, [
+            new ValidationError(dateField, 'Enter a real date')
+        ]);
+    });
+
+
+    it('should reject invalid (non iso format) date', () => {
```
geek-soft, line 115: Remove it.

test/models/chunks/DateSchema.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    it('should accept valid date values', () => {
+        const validationResult = validator.validate({
+            day: '01',
+            month: '01',
+            year: '2020',
+            date: new Date('2020-01-01')
+        });
+        assertValidationErrors(validationResult, []);
+    });
+    it('should accept single digit date values', () => {
```
geek-soft, line 148: Blank line before please.

test/models/chunks/DateSchema.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        ]);
+    });
+
+
+    it('should reject invalid (non-existing) date', () => {
+        const validationResult = validator.validate({
+            day: '33',
+            month: '01',
+            year: '2020',
+            date: '2020-01-33'
```
geek-soft, line 107: Change it to `new Date('2020-01-33')`

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,67 @@
...
+        const startDate: Date = new Date(illness.illnessStart);
+        const day: string = startDate.getDate().toString();
+        const month: string = (startDate.getMonth() + 1).toString();
+        const year: string = startDate.getFullYear().toString();
+
+        return {day, month, year};
+    }
+
+    protected prepareSessionModelPriorSave(appeal: Appeal, value: any): Appeal {
+        const illness: Illness | undefined = appeal.reasons.illness;
```
geek-soft, line 53: Shouldn't that be `appeal.reasons?.illness`?

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+            process.env.ILLNESS_REASON_FEATURE_ENABLED = '1';
+
+            const appealWithReasons = {
+                penaltyIdentifier: {
+                    companyNumber: 'NI000000',
+                    penaltyReference: 'A00000001'
+                },
+                reasons: {
+                    other: {
```
geek-soft, line 60: That should not be in the tests. We do not care about other reasons, lets presume users just started with illness reason.

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,208 @@
...
+                        title: 'I have reasons',
+                        description: 'they are legit'
+                    },
+                    illness: {
+                        illnessStart: '2020-01-01T00:00:00.000Z'
```
geek-soft, line 65: That should be ISO date e.g `2020-07-25`, no time attached.

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    before(() => {
+        initialIllnessReasonFeatureFlag = process.env.ILLNESS_REASON_FEATURE_ENABLED;
+    });
+
+    after(() => {
+        process.env.ILLNESS_REASON_FEATURE_ENABLED = initialIllnessReasonFeatureFlag;
+    });
+
+    const appeal = createDefaultAppeal();
```
geek-soft, line 35: This function creates appeal that has `reasons` pre-populated with other reason is might not what we want. I'd rather have simple object simulating user who just started with illness reason.

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,208 @@
...
+                        .and.to.include(errorSummaryHeading)
+                        .and.to.include(invalidStartDayErrorMessage)
+                        .and.to.include(invalidStartMonthErrorMessage)
+                        .and.to.include(invalidStartYearErrorMessage)
+                        .and.to.not.include(invalidDateErrorMessage)
+                        .and.to.not.include(invalidDateFutureErrorMessage);
+                });
+        });
+
+        it('should return 422 response with rendered error messages when partial start date was submitted',
```
geek-soft, line 126: For the future, I think it is bit to much to check validation in these integration tests. I'd be happy to say it worked after seeing all validation messages after submitting empty form.

You can live it this time but generally not required in the future.

test/controllers/IllnessStartDateController.test.ts
```diff
@@ -0,0 +1,208 @@
...
+                        expect(response.status).to.be.equal(UNPROCESSABLE_ENTITY);
+                        expect(response.text).to.include(pageHeading)
+                            .and.to.include(errorSummaryHeading)
+                            .and.to.include(invalidStartDayErrorMessage)
+                            .and.to.not.include(invalidDateErrorMessage)
+                            .and.to.not.include(invalidDateFutureErrorMessage);
+                    });
+            });
+
+        it('should return 422 response with rendered error message invalid start date (format) was submitted',
```
geek-soft, line 141: I do not think we need that since it is so similar to `should return 422 response with rendered error message invalid start date (non-existing) was submitted`. Having more tests is good but duplicates just increase maintenance overhead.

src/controllers/IllnessStartDateController.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        return {day, month, year};
+    }
+
+    protected prepareSessionModelPriorSave(appeal: Appeal, value: FormBody): Appeal {
+        const illness: Illness | undefined = appeal.reasons?.illness;
+
+        if (illness != null) {
+            illness.illnessStart = value.date.toISOString().split('T')[0];
+            console.log(illness.illnessStart);
```
geek-soft, line 57: No console log please.

test/models/chunks/DateSchema.test.ts
```diff
@@ -0,0 +1,144 @@
...
+            new ValidationError(dateField, 'Enter a real date')
+        ]);
+    });
+
+    it('should reject empty values', () => {
+        const validationResult = validator.validate({
+            day: '',
+            month: '',
+            year: '',
+            date: ''
```
geek-soft, line 61: To be fair it is not real scenario as well. It would be more `--`.

PR #254, opened 2020-07-28
fix: CompanyAuthMiddleware - add redirect to penalty details page when no appeal in session

n-r-b:
### JIRA link
https://companieshouse.atlassian.net/browse/LFA-1295

### Change description
On the CompanyAuthMiddleware, which is applied to most pages in the journey, added a redirect to the penalty details page when there is no appeal data in the session

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/middleware/CompanyAuthMiddleware.ts
+++ b/src/middleware/CompanyAuthMiddleware.ts
@@ -13,6 +13,7 @@ import { CompanyAuthConfig } from 'app/models/CompanyAuthConfig';
 import { Mutable } from 'app/models/Mutable';
 import { SessionStoreConfig } from 'app/models/SessionConfig';
 import { JwtEncryptionService } from 'app/modules/jwt-encryption-service/JwtEncryptionService';
+import { PENALTY_DETAILS_PAGE_URI } from 'app/utils/Paths';
 
 const oathScopePrefix: string = 'https://api.companieshouse.gov.uk/company/';
 
@@ -41,7 +42,8 @@ export class CompanyAuthMiddleware extends BaseMiddleware {
 
         const appeal: Appeal = applicationData.appeal;
         if (!appeal) {
-            return next(new Error('Appeal data not found'));
+            loggerInstance().info(`CompanyAuthMiddleware: Appeal data not found in session, redirecting to ${PENALTY_DETAILS_PAGE_URI}`);
+            return res.redirect(PENALTY_DETAILS_PAGE_URI);
         }
 
         const companyNumber: string = appeal.penaltyIdentifier.companyNumber;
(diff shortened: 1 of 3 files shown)


src/middleware/CompanyAuthMiddleware.ts
```diff
@@ -41,7 +42,8 @@ export class CompanyAuthMiddleware extends BaseMiddleware {
 
         const appeal: Appeal = applicationData.appeal;
         if (!appeal) {
-            return next(new Error('Appeal data not found'));
+            loggerInstance().info(`CompanyAuthMiddleware: Appeal data not found in session, redirecting to ${PENALTY_DETAILS_PAGE_URI}`);
+            return res.redirect(PENALTY_DETAILS_PAGE_URI);
```
akaJSaunders, line 46: If no appeal data is found at this stage, shouldnt be throw an error and fail fast?

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -78,6 +79,34 @@ describe('Company Authentication Middleware', () => {
...
+        const response = createSubstituteOf<Response>();
+        const request = getRequestSubstitute({}, '');
+
+        const startingWithRedirectUrl = (redirect : string) => redirect.startsWith(PENALTY_DETAILS_PAGE_URI);
+
+        await companyAuthMiddleware.handler(request, response, nextFunction);
+        nextFunction.didNotReceive();
+        response.received(1).redirect(Arg.is(startingWithRedirectUrl));
+
+
```
akaJSaunders, line 107:
```suggestion
```

src/middleware/CompanyAuthMiddleware.ts
n-r-b: we previously failed fast and displayed a 500 error page, but then we thought it would be a better user experience if we instead redirected them to the beginning of the journey when no appeal data

test/middleware/CompanyAuthMiddleware.test.ts
```diff
@@ -78,6 +79,33 @@ describe('Company Authentication Middleware', () => {
...
+            companyAuthConfig,
+            sessionStoreForAuthConfig,
+            featureFlag
+        );
+
+        const nextFunction = createSubstituteOf<NextFunction>();
+        const response = createSubstituteOf<Response>();
+        const request = getRequestSubstitute({}, '');
+
+        const startingWithRedirectUrl = (redirect : string) => redirect.startsWith(PENALTY_DETAILS_PAGE_URI);
```
geek-soft, line 101: What is that? Looks like arg matcher. Var names does not help. Since it is used only in one place, why not to inline it to brush the problem under the carpet.

n-r-b: updated this

PR #255, opened 2020-07-28
feat: Continued Illness (Page and Validation)

rhysbarrett:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1440

### Change description

This PR implements and tests the Continued Illness page. It also modifies the Appeal object to allow persistence of data pertaining to an Illness reason.

**Note**: The change to the Appeal object is temporary, as a more thorough update to the Appeal object is necessary (to make the `Other` reason optional and allow multiple reasons) but is outside of the scope of this ticket.

**Second Note**: The page is not currently viewable by manually entering the URL, as it requires session data (the Illness Start Date) which has not yet been implemented. The page is however testable by manipulating the session data.

### Work checklist

### Merge instructions

```
```


src/controllers/ContinuedIllnessController.ts
```diff
@@ -0,0 +1,77 @@
...
+        );
+    }
+
+    protected prepareViewModelFromAppeal(appeal: Appeal): any {
+        const illnessStartDateRaw = appeal.reasons.illness?.illnessStart;
+
+        if (!illnessStartDateRaw) {
+            throw new Error('Illness Start Date expected but not found');
+        } else {
+            const illnessStart = illnessStartDateRaw.toLocaleDateString('en-GB', {
```
geek-soft, line 55: I suppose `illnessStartDateRaw` is a raw string when it is read from session store so I do not think there is `toLocaleDateString` function available in String prototype. I expect this to fail.

src/controllers/ContinuedIllnessController.ts
```diff
@@ -0,0 +1,77 @@
...
+                year: 'numeric', month: 'long', day: 'numeric'
+            });
+            const hint: string = `You told us the illness started on ${illnessStart}`;
+            return { hint };
```
geek-soft, line 59: I'd like to suggest different solution.

I prefer to pass pure data to the template (start date in this instance) and then format it accordingly if needed instead of building UI text in the controllers.

I suggest creating custom Nunjucks filter for date formatting and then using that filter in template as `{{ startDate | formatDate('[format here]') }}`.

This way:
1. wording of the UI remains in template
2. `formatDate` can be reused in many places
3. `formatDate` can be unit tested

src/controllers/ContinuedIllnessController.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    protected prepareSessionModelPriorSave(appeal: Appeal, value: Illness): Appeal {
+        if (appeal.reasons?.illness != null) {
+            appeal.reasons.illness.continuedIllness = value.continuedIllness;
+            if (value.continuedIllness === YesNo.yes) {
```
geek-soft, line 66: Feels unnecessary as values are undefined by default.

src/models/Illness.ts
```diff
@@ -0,0 +1,15 @@
...
+import { IllPerson } from './fields/IllPerson';
+import { YesNo } from './fields/YesNo';
+
+import { Attachment } from 'app/models/Attachment';
+
+export interface Illness {
+    illPerson: IllPerson;
+    otherPerson?: string;
+    illnessStart: Date;
+    continuedIllness: YesNo;
```
geek-soft, line 11: I'd prefer to have only properties which are needed for delivery of the story. Rest could be shared with the team and put under discussion if not done yet.

src/models/fields/IllPerson.ts
```diff
@@ -0,0 +1,7 @@
+export enum IllPerson {
```
geek-soft, line 1: Same here, it should not be added by this PR.

src/views/continued-illness.njk
```diff
@@ -0,0 +1,64 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      {{
+        govukErrorSummary ({
+          titleText: 'There is a problem with the details you gave us',
+          errorList: validationResult.errors
+        }) if validationResult and validationResult.errors.length > 0
+      }}
+
+      <form method="post" id="continued-illness-form">
```
geek-soft, line 22: I wonder whether we need form ID. Why did you add it?

src/controllers/ContinuedIllnessController.ts
```diff
@@ -0,0 +1,77 @@
...
+    },
+    actions: (_: boolean) => {
+        return {
+            continue:'action=continue'
+        };
+    }
+};
+
+@controller(CONTINUED_ILLNESS_PAGE_URI, IllnessReasonFeatureMiddleware, SessionMiddleware, AuthMiddleware)
+export class ContinuedIllnessController extends BaseController<any> {
```
geek-soft, line 34: That `<any>` should be replaced with the type that corresponds to data structure behind HTML from the template. In other wards it should be the data structure of `request.body`.

src/controllers/ContinuedIllnessController.ts
rhysbarrett: True, however this logic covers the following scenario:

1.  User enters illness journey and ticks "Is the person still ill?" as No
2.  User is then asked to enter an illness end date
3.  User gets near the end of the journey, decides they've made a mistake, and navigates back to this page
4.  User then instead ticks Yes for "Is the person still ill"

In this scenario, without resetting `illnessEnd` to `undefined`, the illness would have conflicting values:
* `continuedIllness`: Yes
* `illnessEnd`: <Some date in the past>

This wouldn't make sense from a business perspective, so we must nullify `illnessEnd` whenever `Yes` is chosen on this page.

src/controllers/ContinuedIllnessController.ts
```diff
@@ -0,0 +1,77 @@
...
+        } else {
+            const illnessStart = illnessStartDateRaw.toLocaleDateString('en-GB', {
+                year: 'numeric', month: 'long', day: 'numeric'
+            });
+            const hint: string = `You told us the illness started on ${illnessStart}`;
+            return { hint };
+        }
+    }
+
+    protected prepareSessionModelPriorSave(appeal: Appeal, value: Illness): Appeal {
```
geek-soft, line 63: Value argument seems to have wrong type. Is should be generic type named `FORM` expected by `BaseController<FORM>`. As I mentioned earlier it should corrections to data structure of the form handled by POST handler.

src/controllers/ContinuedIllnessController.ts
rhysbarrett: I should confirm this with Natalie. My feeling is that Date validation will be far easier if the date is in a Date format, rather than a string.

src/controllers/ContinuedIllnessController.ts
```diff
@@ -0,0 +1,77 @@
...
+            template,
+            navigation,
+            new FormValidator(schema)
+        );
+    }
+
+    protected prepareViewModelFromAppeal(appeal: Appeal): any {
+        const illnessStartDateRaw = appeal.reasons.illness?.illnessStart;
+
+        if (!illnessStartDateRaw) {
```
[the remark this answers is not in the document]
rhysbarrett: Initially entered this to surpass the type-checking (was throwing an error because the object may be undefined) but I've now found another way around this which doesn't involve error handling.

test/controllers/ContinuedIllnessController.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    it('should show hint containing illness start date, and radio buttons for Yes and No on GET', async () => {
+        const  applicationData: Partial<ApplicationData> = {
+            appeal: {
+                penaltyIdentifier: {
+                    companyNumber: 'NI000000'
+                },
+                reasons: {
+                    illness: {
+                        illnessStart: new Date(2020, 4, 1)
```
geek-soft, line 21: Test simulates date in `Date` format but without custom deserialisation `String` type will be used instead. It is because node-session-handler does simple `JSON.stringify()` to serialise and `JSON.parse` to deserialise.

In other words test simulates scenario that will never happen.

test/controllers/ContinuedIllnessController.test.ts
```diff
@@ -0,0 +1,60 @@
...
+            expect(res.text).to.include('type="radio"');
+            expect(res.text).to.include('value="yes"');
+            expect(res.text).to.include('value="no"');
+            const radioCount = (res.text.match(/type="radio"/g) || []).length;
+            expect(radioCount).to.equal(2);
+        });
+
+    });
+
+
```
geek-soft, line 42: Unnecessary blank line. Let's keep no more than one between tests.

src/controllers/ContinuedIllnessController.ts
rhysbarrett: Do we have any instances of custom Nunjucks filters elsewhere in the code? I did this to avoid meddling with the BaseController but can of course look into this. I originally wanted to only pass the formatted date to the template, but Nunjucks doesn't support nested variables (this occurs within a `govukRadios` macro).

src/models/Illness.ts
rhysbarrett: Noted and ammended.

test/controllers/ContinuedIllnessController.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    it('should show an error if no option is selected on POST', async () => {
```
geek-soft, line 43: Shall we add another one when validation passed to expect page reload?

src/models/fields/IllPerson.ts
rhysbarrett: Removed as requested.

src/views/continued-illness.njk
rhysbarrett: Good question. I thought I'd followed the GOV.UK Design System example but apparently not - this is an artefact of referencing the Select The Penalty page. I'll remove.

src/controllers/ContinuedIllnessController.ts
geek-soft: Nope, but here is an example from project I previously worked for: https://github.com/hmcts/cmc-citizen-frontend/blob/master/src/main/modules/nunjucks/filters/dateFilter.ts#L18

geek-soft: Nunjucks filters have no dependency on BaseController.

In nutshell you would need to:
1. implement filter method
2. register filter with templating engine
3. use registered filter in the code

If we have one date format you can make it simple / argument-less for now so it always formats date into one format.

src/controllers/ContinuedIllnessController.ts
rhysbarrett: The compiler doesn't seem to like this, and I can't see any instances of this in the codebase. Perhaps it should be Illness instead?
Screenshot 2020-07-29 at 10 40 38

src/controllers/ContinuedIllnessController.ts
geek-soft: Good call, forgot about it. Ignore my previous comment in that case.

src/controllers/ContinuedIllnessController.ts
rhysbarrett: Again I can't see any instances of this in the codebase, and it doesn't compile unless I change the controller definition to:
`export class ContinuedIllnessController<FORM> extends BaseController<FORM> {`

No other top-level controllers in the codebase are generic in this way.

Screenshot 2020-07-29 at 10 44 58

src/controllers/ContinuedIllnessController.ts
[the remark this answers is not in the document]
geek-soft: `appeal.reasons.illness!.illnessStart`? Exclamation mark will silence type checking telling type checker that in this scenario start date will always be there.

test/controllers/ContinuedIllnessController.test.ts
rhysbarrett: Ah. So you're saying we _have_ to store the dates as strings? This really doesn't sound like a good idea from a data design perspective.

test/controllers/ContinuedIllnessController.test.ts
rhysbarrett: Well spotted. Removed.

src/controllers/ContinuedIllnessController.ts
geek-soft: If you guys want to use Date you need to have custom deserialisation logic cause in Redis Date will be stored as string.

src/views/continued-illness.njk
geek-soft: I generally do not like having things that are useless cause their existence might suggest otherwise.

test/controllers/ContinuedIllnessController.test.ts
rhysbarrett: My feeling is that adding a test case for a temporary effect (rather than desired feature) wouldn't be a good idea but happy to do this if you feel it's necessary.

src/controllers/ContinuedIllnessController.ts
rhysbarrett: *cries*

src/controllers/ContinuedIllnessController.ts
geek-soft: It should be data type behind HTML form! See below how it is used:

```
prepareSessionModelPriorSave(applicationData.appeal || {}, request.body);
```

src/controllers/ContinuedIllnessController.ts
rhysbarrett: Okay, thanks for the explanation. I'll get on that - though I'll get in touch with Natalie to see what datestring format she's using first.

src/views/continued-illness.njk
rhysbarrett: Absolutely.

src/controllers/ContinuedIllnessController.ts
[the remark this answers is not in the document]
rhysbarrett: Great minds think alike, that's exactly what I've done.

src/controllers/ContinuedIllnessController.ts
geek-soft: Maybe others kept ignoring it when I wasn't looking :( `OtherReasonController` has it, although type isn't 100% correct cause it has attachments in it despite attachment wasn't in `request.body`. You can define form type locally in the controller class which is way better than ANY.

test/controllers/ContinuedIllnessController.test.ts
geek-soft: I do not, I am saying after read from sessions store value will be in string type. Type definitions and actual type might be completely different at runtime, it is developers job to make sure they mach 100%.

test/controllers/ContinuedIllnessController.test.ts
geek-soft: Maybe that test should only check that 200 response is given? In a way that would test that page validation can passes.

Anyway just suggestion, I do not feel strong about that, can be added when navigation is implemented. Will live it up to your judgement.

rhysbarrett: 200 response check sounds like a good compromise, I'll implement that.

src/controllers/ContinuedIllnessController.ts
rhysbarrett: Discussion continued on Slack. Added more specific Interface definition to controller and now using this type in definition. Marking as resolved.

src/controllers/ContinuedIllnessController.ts
rhysbarrett: Implemented as discussed over Slack. Marking as resolved.

src/modules/nunjucks/DateFilter.ts
```diff
@@ -0,0 +1,27 @@
+export function dateFilter(value: string ): string {
+    // tslint:disable-next-line: prefer-const
+    let [year, month, day] = value.split('-').map(i => {
```
geek-soft, line 3: That could be declared as `const`

src/modules/nunjucks/DateFilter.ts
```diff
@@ -0,0 +1,27 @@
+export function dateFilter(value: string ): string {
```
geek-soft, line 1: Unnecessary whitespace before closing bracket.

src/modules/nunjucks/DateFilter.ts
geek-soft, line 3: Shouldn't formatting function be more graceful so that if value is not present it does not blow up? I'd expect this to return undefined or null if input value is undefined or null.

src/modules/nunjucks/DateFilter.ts
geek-soft: There should also be validation if string value does not conform ISO format. Regex could be used.

src/modules/nunjucks/DateFilter.ts
```diff
@@ -0,0 +1,27 @@
...
+    // tslint:disable-next-line: prefer-const
+    let [year, month, day] = value.split('-').map(i => {
+        return parseInt(i, 10);
+    });
+
+    if (!year || !month || !day || month > 11 || month < 0) {
+        throw new Error('Input should be an ISO-formatted date string');
+    }
+
+    const months = [
```
geek-soft, line 11: This feels unnecessary. Wouldn't it be better to use date formatting library of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat for example?

src/utils/ExpressConfigFunctionFactory.ts
```diff
@@ -30,6 +31,8 @@ export const createExpressConfigFunction = (directory: string) => (app: express.
         express: app,
     });
 
+    nunjucksEnv.addFilter('govukDate', dateFilter);
```
geek-soft, line 34: Not a big fan of this approach. What gov UK date really is? Is there such thing? I'd just call it `date`. I am curious your reason behind that choice.

test/controllers/ContinuedIllnessController.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    it('should show hint containing illness start date, and radio buttons for Yes and No on GET', async () => {
+        const  applicationData: Partial<ApplicationData> = {
+            appeal: {
+                penaltyIdentifier: {
+                    companyNumber: 'NI000000'
+                },
+                reasons: {
+                    illness: {
+                        illnessStart: '2020-5-1'
```
geek-soft, line 21: I'd expect this to be correct ISO date in YYYY-MM-DD format. Otherwise that would likely need to be reformatted on submission to Appeals API. Can you ask @n-r-b to make sure we store correct ISO string?

src/utils/ExpressConfigFunctionFactory.ts
rhysbarrett: It was a temporary change to debug an issue with the pipeline, which I then forgot to revert. I'll put it back to `date`.

src/modules/nunjucks/DateFilter.ts
rhysbarrett: I attempted something similar to what you're suggesting but it lead to inconsistent output between my local machine and the pipeline. I'll try one of these just in case.

src/modules/nunjucks/DateFilter.ts
rhysbarrett: I would prefer this to fail loudly than silently, as it is a function which is only used by other programmer-defined functions, rather than as a direct result of user input. If this filter breaks, the person writing the code using it should know about it.

Unless quiet failing is preferred in TypeScript, in which case I'd be happy to be more graceful.

src/modules/nunjucks/DateFilter.ts
geek-soft: Isn't that a matter of locale set on the machine. Maybe pipelines use something different. Have you tried setting locale in constructor .e.g. `new Intl.DateTimeFormat('en-GB')`?

src/modules/nunjucks/DateFilter.ts
rhysbarrett: Regex would definitely be neater, thanks for the suggestion.

src/modules/nunjucks/DateFilter.ts
geek-soft: I generally prefer to fail fast and loud. I thought maybe allowing undefined / null values would be convenient for developers so no 'if' statements needs to be added in the code but you might be right, I do not find good use case for UI that would print empty string if date value wasn't defined. In that case ignore my initial comment.

rhysbarrett: Ah, yes. The let is an artefact of when I was decrementing the `month` value directly, instead of simply subtracting two upon reference as I am now. I'll change to const.

src/modules/nunjucks/DateFilter.ts
rhysbarrett: I was originally using `Date.toLocaleDateString('en-GB', options)` so was very confused when the pipeline gave a different output. My assumption was that by specifying the locale, the output would be consistent. When I realised this wasn't the case I hard-coded the solution you can see now. I'll try it with the Intl.DateTimeFormat as suggested and see if the output is still inconsistent.

test/controllers/ContinuedIllnessController.test.ts
rhysbarrett: Noted and changed. I'll confirm with Natalie tomorrow.

src/modules/nunjucks/DateFilter.ts
rhysbarrett: Nope, this still isn't working. See screenshot. I have no idea why specifying the locale doesn't lock down the format - it's infuriating.
Screenshot 2020-07-29 at 17 34 43

geek-soft: Alright, due to time constraints I am happy to accept it for now.

PR #259, opened 2020-07-30
refactor: Rename environment variable file, remove support for multiple files

rhysbarrett:
### JIRA link

N/A

### Change description

This PR renames `.env.vagrant` as `.env.test` and removes support for loading of multiple environment files.

### Work checklist

### Merge instructions

```
```


Dockerfile
```diff
@@ -27,7 +27,7 @@ WORKDIR /app
 
 COPY --from=prod-deps-image /build/node_modules/ ./node_modules
 COPY --from=build-image /build/dist/ ./dist
-COPY .env.vagrant ./
+COPY .env.test ./
```
geek-soft, line 30: To be fair it should not be copied if this is just test file.

src/utils/ConfigLoader.ts
```diff
@@ -1,11 +1,12 @@
...
+
+const DEFAULT_ENV_FILE = `.env`;
 
 export const loadEnvironmentVariablesFromFiles = () => {
-    dotenv.config({ path: DEFAULT_ENV_FILE });
-    if (process.env.NODE_ENV) {
-        const envFilePath = `${__dirname}/../../.env.${process.env.NODE_ENV}`;
-        dotenv.config({ path: envFilePath });
-    }
+    const env = getEnv('NODE_ENV');
```
geek-soft, line 8: I believe we should simplify it so only `.env.[env]` is loaded.

Dockerfile
rhysbarrett: Oh okay, cool. I'll remove.

src/utils/ConfigLoader.ts
rhysbarrett: Okay, and we allow it to fail loudly if the environment variable isn't present?

src/utils/ConfigLoader.ts
```diff
@@ -1,11 +1,9 @@
...
-const DEFAULT_ENV_FILE = `${__dirname}/../../.env`;
+import { getEnvOrThrow } from 'app/utils/EnvironmentUtils';
 
 export const loadEnvironmentVariablesFromFiles = () => {
-    dotenv.config({ path: DEFAULT_ENV_FILE });
-    if (process.env.NODE_ENV) {
-        const envFilePath = `${__dirname}/../../.env.${process.env.NODE_ENV}`;
-        dotenv.config({ path: envFilePath });
-    }
+    const env = getEnvOrThrow('NODE_ENV');
```
geek-soft, line 6: A `NODE_ENV` env is optional, using `getEnvOrThrow` will break application in cloud environments. Use more graceful option.

rhysbarrett: This seems to conflict with your previous comment: https://github.com/companieshouse/lfp-appeals-frontend/pull/259#discussion_r463078369

Either we assume NODE_ENV is present and break the application when it isn't, or we need a default/fallback file.

geek-soft: In previous comment I meant to stop loading `.env` and load `.env.[env]` but only when `NODE_ENV` is present. We always had conditional use of `NODE_ENV` and we should not change it.

PR #260, opened 2020-07-31
feat: Transitions from Illness Start Date page

rhysbarrett:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1444

### Change description

This PR adds navigation to the Illness Start Date page:

- Back Button: Leads to the Who Was Ill page
- Continue: Now leads to the Continued Illness page

**Note:** A temporary placeholder page has been created for the Who Was Ill page to allow automated testing of this feature before the implementation of the Who Was Ill page is complete.

**Note:** This feature is dependent on LFA 1439 ( PR #253 ) and as such should not be merged prior.

### Work checklist

### Merge instructions

```
```


src/controllers/WhoWasIllController.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { BaseAsyncHttpController } from 'app/controllers/BaseAsyncHttpController';
+import { FeatureToggleMiddleware } from 'app/middleware/FeatureToggleMiddleware';
+import { Feature } from 'app/utils/Feature';
+import { WHO_WAS_ILL_PAGE_URI } from 'app/utils/Paths';
+
+const template = 'illness/who-was-ill';
+
+@controller(WHO_WAS_ILL_PAGE_URI, FeatureToggleMiddleware(Feature.ILLNESS_REASON), SessionMiddleware)
+export class WhoWasIllController extends BaseAsyncHttpController {
```
akaJSaunders, line 12: Why are we not using base controller here?

rhysbarrett: Just wanted to knock a quick and easy controller up with minimal code, as it isn't included in this story. I'm hoping to pick up the relevant story next so will implement the controller properly (and using BaseController) then.

src/controllers/WhoWasIllController.ts
```diff
@@ -0,0 +1,18 @@
...
+import { controller, httpGet } from 'inversify-express-utils';
+
+import { BaseAsyncHttpController } from 'app/controllers/BaseAsyncHttpController';
+import { FeatureToggleMiddleware } from 'app/middleware/FeatureToggleMiddleware';
+import { Feature } from 'app/utils/Feature';
+import { WHO_WAS_ILL_PAGE_URI } from 'app/utils/Paths';
+
+const template = 'illness/who-was-ill';
+
+@controller(WHO_WAS_ILL_PAGE_URI, FeatureToggleMiddleware(Feature.ILLNESS_REASON), SessionMiddleware)
```
akaJSaunders, line 11: This feels a bit out of scope, I don't remember from the standup if this is what we agreed to do. Ideally we would have that page ready before implementing this ticket. Can we not link the user to the URL even if a page doesnt exist?

src/controllers/ContinuedIllnessController.ts
```diff
@@ -14,7 +14,7 @@ import { Feature } from 'app/utils/Feature';
 import { CONTINUED_ILLNESS_PAGE_URI } from 'app/utils/Paths';
 import { Navigation } from 'app/utils/navigation/navigation';
 
-const template = 'continued-illness';
+const template = 'illness/continued-illness';
```
akaJSaunders, line 17: the template is not the same thing as the URL. this should be the name of the file in the view directory. 

```suggestion
const template = 'continued-illness';
```

src/controllers/WhoWasIllController.ts
rhysbarrett: We _can_, and it wouldn't affect local unit tests. However, it would lead to a 404 status instead of a 200, which could impact the automated tests. Since you've been drafted into Testing, you'd know about this, so if you feel either a) the tests would be unaffected or b) they _would_ be affected but you'd be happy to update the test once the Who Was Ill page is implemented, then I'm more than happy to get rid of this placeholder page.

src/controllers/ContinuedIllnessController.ts
rhysbarrett: Indeed. However, Natalie updated the structure of the `/views/` directory, presumably to neaten the whole thing up a bit, so I've followed suit.

src/controllers/WhoWasIllController.ts
```diff
@@ -0,0 +1,18 @@
+import { SessionMiddleware } from 'ch-node-session-handler';
+import { controller, httpGet } from 'inversify-express-utils';
+
+import { BaseAsyncHttpController } from 'app/controllers/BaseAsyncHttpController';
+import { FeatureToggleMiddleware } from 'app/middleware/FeatureToggleMiddleware';
+import { Feature } from 'app/utils/Feature';
+import { WHO_WAS_ILL_PAGE_URI } from 'app/utils/Paths';
+
+const template = 'illness/who-was-ill';
```
akaJSaunders, line 9: Similar to before, the template is not the same thing as the URL. this should be the name of the file in the view directory.

```suggestion
const template = 'who-was-ill';
```

rhysbarrett: Similiar to before - it is :)

src/controllers/ContinuedIllnessController.ts
akaJSaunders: oh fancy

src/controllers/WhoWasIllController.ts
akaJSaunders: oh fancy

src/controllers/ContinuedIllnessController.ts
rhysbarrett: Indeed! Might be nice to do the same to the controllers directory at a later stage, as that thing's getting looooong. But that's a story for another time.

src/controllers/WhoWasIllController.ts
geek-soft, line 12: I am not big fan of these names looking like a question. Name that person the way it would be named in our data structures and then rename controller to use same noun.

src/controllers/WhoWasIllController.ts
geek-soft, line 11: Why is there session middleware applied to it?

src/controllers/WhoWasIllController.ts
rhysbarrett: Completely agree. I'll fix that.

src/controllers/WhoWasIllController.ts
rhysbarrett: Good question, I'll remove.

PR #261, opened 2020-08-03
bugfix: illness start date session reading bug

akaJSaunders:
### Change description

When the page is reloaded after you enter a date, the field is prepopulated with the session as it should ... except that:
for some months, the value of the day, is decremented by one.I dont think this is a problem about how we save the date to session, I think its to do with how we read it.

most likely linked to how new Date() works.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/IllnessStartDateController.ts
+++ b/src/controllers/IllnessStartDateController.ts
@@ -41,10 +41,14 @@ export class IllnessStartDateController extends BaseController<FormBody> {
         if (!illness) {
             return {};
         }
-        const startDate: Date = new Date(illness.illnessStart);
-        const day: string = startDate.getDate().toString();
-        const month: string = (startDate.getMonth() + 1).toString();
-        const year: string = startDate.getFullYear().toString();
+
+        console.log(illness.illnessStart);
+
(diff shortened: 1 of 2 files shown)


src/controllers/IllnessStartDateController.ts
```diff
@@ -41,10 +41,12 @@ export class IllnessStartDateController extends BaseController<FormBody> {
...
-        const startDate: Date = new Date(illness.illnessStart);
-        const day: string = startDate.getDate().toString();
-        const month: string = (startDate.getMonth() + 1).toString();
-        const year: string = startDate.getFullYear().toString();
+
+        const dateIngredients = illness.illnessStart.split('-', 3);
+
+        const year: string = dateIngredients[0];
+        const month: string = dateIngredients[1];
+        const day: string = dateIngredients[2];
```
rhysbarrett, line 49: An even cleaner way of doing this would be:
`const [year, month, day] = value.split('-', 3);`

rhysbarrett:
Should the commit start with `fix:` rather than `bugfix:` for the semantic versioning to work?

PR #262, opened 2020-08-03
fix: time zone bug

akaJSaunders:
### Work checklist

### Merge instructions

```
```


src/utils/DateFormatter.ts
```diff
@@ -0,0 +1,12 @@
+export const dateToString = (date: Date): string => {
+
+    if(isNaN(date.getDate())){
```
geek-soft, line 3: Spaces around curly brackets would be great.

test/utils/DateFormatter.test.ts
```diff
@@ -0,0 +1,57 @@
+import { expect } from 'chai';
+
+import { dateToString } from 'app/utils/DateFormatter';
+
+describe('DateFormatter', () => {
+
+    it('should throw invalid date error', () =>{
+
+        const date: Date = new Date('nonsense date value');
```
geek-soft, line 9: Love that string value.

test/utils/DateFormatter.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+
+    it('should return YYYY-MM-DD format for a local date', () =>{
+
+        const date: Date = new Date('2020-10-30T23:00:00.000');
+        expect(dateToString(date)).to.equal('2020-10-30');
+
+    });
+
+    it('should return YYYY-MM-DD format for a UTC date', () =>{
```
geek-soft, line 21: Not sure what you testing here. Dates are the same cause in October GMT is being used in UK which has the same offset as UTC.

test/utils/DateFormatter.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        const date: Date = new Date('2020-10-30T23:00:00.000z');
+        expect(dateToString(date)).to.equal('2020-10-30');
+
+    });
+
+    it('should return incremented month value [0-11] by one for correct string representation [1-12]', () =>{
```
geek-soft, line 28: That sounds super low level and it only applies to dates created from individual date components. It does not look good to me.

test/utils/DateFormatter.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        const date: Date = new Date();
+        date.setDate(30);
+        date.setMonth(4);
+        date.setFullYear(2020);
+        expect(dateToString(date)).to.equal('2020-05-30');
+
+    });
+
+    it('should return a padded day format for single digit days', () =>{
```
geek-soft, line 48: Last two test do the same, just for different date components, shouldn't these two tests have its name is worded in similar way?

test/utils/DateFormatter.test.ts
akaJSaunders: Its a small little detail that probably isnt worth testing, but notice the "z" at the end of the string:
`new Date('2020-10-30T23:00:00.000z'`

this denotes that the timezone is zero utc offset. I guess in this case im making sure that format works, but I agree thats its not necessary

test/utils/DateFormatter.test.ts
akaJSaunders: Sounds like sarcasm

src/utils/DateFormatter.ts
```diff
@@ -0,0 +1,12 @@
+export const dateToString = (date: Date): string => {
```
geek-soft, line 1: The whole point of having that function is fixing issue around timezone differences although it does not say that either in function name or in JS doc. Make it more obvious that formatted date is in local time.

test/utils/DateFormatter.test.ts
akaJSaunders: this `dateToString` function could be called from other locations in the future, I want to make sure to have a test that covers this case. Its a small but quite dangerous little detail

akaJSaunders: renamed it to : 
`'should return correct month number for a given month value'`

test/utils/DateFormatter.test.ts
```diff
@@ -0,0 +1,50 @@
...
+describe('DateFormatter', () => {
+
+    it('should throw invalid date error', () =>{
+
+        const date: Date = new Date('nonsense date value');
+        expect(() => dateToLocalTimeString(date)).throws('DateFormatter - Invalid date');
+
+    });
+
+    it('should return YYYY-MM-DD format for a local date', () =>{
```
geek-soft, line 14: What about scenario where timezones differ? The main purpose of creating that function is not really tested.

test/utils/DateFormatter.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        const date: Date = new Date('2020-10-30T23:00:00.000');
+        expect(dateToLocalTimeString(date)).to.equal('2020-10-30');
+
+    });
+
+    it('should return correct month number for a given month value', () =>{
```
geek-soft, line 21: I do not really understand what that test scenario means. What does `correct month number for a given month value` really mean?

akaJSaunders: when you set values you can set:

- Dates to any value from 1 to 31
- Months to any value from 0 to 11. However in string format we want to make sure to add 1 to that value so that we store and display months from 1-12

test/utils/DateFormatter.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    });
+
+    it('should return YYYY-MM-DD format for a date', () =>{
+
+        const date: Date = new Date('2020-10-30T23:00:00.000');
+        expect(dateToLocalTimeString(date)).to.equal('2020-10-30');
+
+    });
+
+    it('should return string for local date rather than UTC', () =>{
```
geek-soft, line 21: How node knows what the local time is? Aren't these tests going to fail when time zone changed into GMT later this year?

PR #263, opened 2020-08-03
feat: "Who Was Ill?" page and validation

rhysbarrett:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1438

### Change description

This commit implements the "Who Was Ill?" page, along with validation. See screenshots below.

**Accessibility Note:** The Conditional Input causes AXE to throw a critical accessibility error. [This is a known and accepted issue within the GOV.UK Frontend system](https://github.com/alphagov/govuk-frontend/issues/1280#issuecomment-509588851).

image
image
image

### Work checklist

### Merge instructions

We are committed to keeping commit history clean, consistent and linear. To acheive this commit should be structured as follows:

```
```


src/controllers/IllPersonController.ts
```diff
@@ -1,17 +1,89 @@
...
+            return {};
+        }
+
+        return {
+            illPerson: illness.illPerson,
+            otherPerson: illness.otherPerson
+        };
+    }
+
+    protected prepareSessionModelPriorSave(appeal: Appeal, value: any): Appeal {
```
akaJSaunders, line 63: value is of type FormBody correct?
```suggestion
    protected prepareSessionModelPriorSave(appeal: Appeal, value: FormBody): Appeal {
```

src/controllers/IllPersonController.ts
```diff
@@ -1,17 +1,89 @@
...
+
+        const dummyOther: OtherReason = {
+            title: 'Dummy Reason',
+            description: 'The current Appeal data model requires an Other-type reason, ' +
+                'and reworking the data model is outside of the scope of this feature. ' +
+                'Until the Appeal object has been remodelled, this dummy reason must ' +
+                'be included.'
+        };
```
akaJSaunders, line 71: I see why you did this, but I don't like it. Is there maybe another way of getting around it?

akaJSaunders: `const emptyOther = {} as OtherReason`

would this pehaps work?

src/controllers/IllPersonController.ts
```diff
@@ -1,17 +1,89 @@
...
+                illness: value,
+                other: dummyOther
+            };
+        }
 
-    @httpGet('')
-    public async redirectView (): Promise<void> {
-        return this.render(template);
+        loggerInstance()
+            .debug(`${IllPersonController.name} - prepareSessionModelPriorSave: ${JSON.stringify(appeal)}`);
```
akaJSaunders, line 86: Im not sure we need this log, considering we don't do this for every controller, id say we can remove it.

src/views/illness/ill-person.njk
```diff
@@ -1,16 +1,95 @@
...
+            id: "other-person",
+            name: "otherPerson",
+            type: "text",
+            spellcheck: false,
+            classes: "govuk-!-width-one-third",
+            label: {
+              text: "Their relationship to the company"
+            },
+            errorMessage: validationResult.getErrorForField('otherPerson') if validationResult
+          }) }}
```
akaJSaunders, line 36: could you reformat this govukInput so that its similar to lines 17-22

src/views/illness/ill-person.njk
```diff
@@ -1,16 +1,95 @@
...
+          errorList: validationResult.errors
+        }) if validationResult and validationResult.errors.length > 0
+      }}
+
+      <form method="post">
+        {% set otherPersonHtml %}
+          {{ govukInput({
+            id: "other-person",
+            name: "otherPerson",
+            type: "text",
```
akaJSaunders, line 29: within this components we usually use single quotes, could you fix this through the template. Html elements can remain with double quotes

src/controllers/IllPersonController.ts
rhysbarrett: I also don't like it. Unfortunately this is the only way to get around the issue until we rework the data structure, which is going to be a huge task. I cleared this with Damian on a previous ticket.

src/controllers/IllPersonController.ts
rhysbarrett: We do it in every controller where we use `prepareSessionModelPriorSave`.

test/controllers/IllPersonController.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    const applicationData: Partial<ApplicationData> = {
+        appeal: {
+            penaltyIdentifier: {
+                companyNumber: 'NI000000'
+            }
+        } as Appeal,
+        navigation: { permissions: [ILL_PERSON_PAGE_URI] }
+    };
+
+    it('should show radio buttons for available Ill Person options on GET', async () => {
```
akaJSaunders, line 20: something we have done in a few of the tests which im quite fond of is separate GETS and POST tests.

What I mean exactly is: 

- Within the main describe, you can add one for GET and another one for POST. 
```
describe('IllPersonController', () => {
    describe('GET request', () => {}
    describe('POST request', () => {}
}
```
You can then remove the "GET" and "POST" from the names of the tests. Keeps it organized and easier to read.

test/controllers/IllPersonController.test.ts
```diff
@@ -0,0 +1,63 @@
...
+            expect(res.text).to.include('class="govuk-radios__conditional govuk-radios__conditional--hidden"');
+            expect(res.text).to.include('name="otherPerson"');
+        });
+    });
+
+    it('should show an error if no person is selected on POST', async () => {
+        const app = createApp(applicationData);
+
+        await request(app).post(ILL_PERSON_PAGE_URI).expect(res => {
+            expect(res.status).to.equal(422);
```
akaJSaunders, line 48: Could you remove these status code numbers, and import the code from `'http-status-codes'`;

For example in OtherReasonController.ts
```
import { MOVED_TEMPORARILY, OK, UNPROCESSABLE_ENTITY } from 'http-status-codes';
...
expect(response.status).to.be.equal(UNPROCESSABLE_ENTITY);
```

test/controllers/IllPersonController.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    it('should hide the Other Person conditional input by default', async () => {
+        const app = createApp(applicationData);
+
+        await request(app).get(ILL_PERSON_PAGE_URI).expect(res => {
+            expect(res.text).to.include('class="govuk-radios__conditional govuk-radios__conditional--hidden"');
+            expect(res.text).to.include('name="otherPerson"');
+        });
+    });
+
+    it('should show an error if no person is selected on POST', async () => {
```
akaJSaunders, line 44: I'd rename this test to 'should show validation error ....'

src/controllers/IllPersonController.ts
rhysbarrett: Well spotted, thanks.

test/controllers/IllPersonController.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        const app = createApp(applicationData);
+
+        await request(app).post(ILL_PERSON_PAGE_URI).expect(res => {
+            expect(res.status).to.equal(422);
+            expect(res.text).to.contain('You must select a person');
+        });
+    });
+
+
+    it('should show an error if "Other" is chosen but not specified on POST', async () => {
```
akaJSaunders, line 54: I'd rename this test to 'should show validation error ....'

src/views/illness/ill-person.njk
rhysbarrett: Good spot, fixed.

src/views/illness/ill-person.njk
rhysbarrett: Certainly - fixed.

test/models/IllPersonSchema.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+const validator = new SchemaValidator(schema);
+const emptySelectionErrorMessage = 'You must select a person';
+const emptyOtherPersonErrorMessage = 'You must tell us more information';
+
+describe('IllPerson schema', () => {
+
+    describe('should reject', () => {
+
+        it('empty object with only one error', () => {
```
akaJSaunders, line 16: The name of these tests doesn't make much sense within the context of the code because of the `it() `functon. Id start it these names with `should....`

I would also rename the describe they are all within. Id be happy with just `rejects`

test/controllers/IllPersonController.test.ts
rhysbarrett: Ah yes, I've seen this in some cases but not others, the inconsistency is upsetting. I prefer this so I'll switch to it. Might be nice to refactor all of the controller tests to use this pattern in future.

test/models/IllPersonSchema.test.ts
akaJSaunders: If you could fix that through this code that would be great <3

src/controllers/IllPersonController.ts
akaJSaunders: lol my bad

test/controllers/IllPersonController.test.ts
rhysbarrett: Completely agree, consider it done.

test/controllers/IllPersonController.test.ts
rhysbarrett: Agreed and fixed.

test/models/IllPersonSchema.test.ts
rhysbarrett: I see what you mean. However these changes would lead to the following output when running the tests:
```
IllPerson schema
    rejects
        should reject empty object with only one error
```
for example. Are we prioritising readability of code or test output?

test/controllers/IllPersonController.test.ts
rhysbarrett: Agreed and fixed.

src/controllers/IllPersonController.ts
akaJSaunders: did you try my suggestion?
`const emptyOther = {} as OtherReason`

src/models/Illness.ts
```diff
@@ -1,7 +1,10 @@
 
-import { YesNo } from './fields/YesNo';
+import { IllPerson } from 'app/models/fields/IllPerson';
+import { YesNo } from 'app/models/fields/YesNo';
```
akaJSaunders, line 3: Why does we need YesNo here?

src/controllers/IllPersonController.ts
rhysbarrett: I originally did try setting `other: {}` but that didn't work. Turns out adding the `as OtherReason` to the end of it makes it work. TypeScript is _wild_. Thank you.

src/models/fields/IllPerson.schema.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const schema = Joi.object({
+    illPerson: Joi.string()
+        .required()
+        .valid(...Object.values(IllPerson).filter(x => typeof x === 'string'))
+        .messages({
+            'any.required': emptySelectionErrorMessage,
+            'any.only': emptySelectionErrorMessage,
+            'string.base': emptySelectionErrorMessage,
+            'string.empty': emptySelectionErrorMessage
```
akaJSaunders, line 16: I noticed you are checking the type here (any and string). Isn't this for radio buttons? Why are we checking string.base, string.empty

src/models/Illness.ts
rhysbarrett: It just makes mapping from the user input easier (than converting a yes/no input to a boolean.. aye, it's likely trivial, but if we already have an enum defined then why not use it?) and will presumably make it slightly easier when presenting the data back to the user.

src/models/fields/IllPerson.schema.ts
rhysbarrett: People can edit radio buttons through the developer console. But... come to think of it, only a preset list of values are valid anyway, so these two lines are indeed redundant. Nice spot.

src/models/fields/IllPerson.ts
```diff
@@ -0,0 +1,7 @@
+export enum IllPerson {
+    director = 'director',
+    accountant = 'accountant',
+    family = 'family',
+    employee = 'employee',
+    otherPerson = 'otherPerson'
```
akaJSaunders, line 6: why did you choose to use otherPerson as the name for the field, rather than “someoneElse”. It matches better with what is represented on the page

src/views/illness/ill-person.njk
```diff
@@ -1,16 +1,95 @@
...
 {% endblock %}
 
 {% block content %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <h1 class="govuk-heading-xl">
-        Who was ill?
-      </h1>
-      <p class="govuk-body">This page is currently in development.</p>
+
```
akaJSaunders, line 16:
```suggestion
```

src/models/fields/IllPerson.ts
rhysbarrett: Just following the prototype, but happy to change if you think it's needed. In fact this would help distinguish between the `illPerson` value "otherPerson", and the field called "otherPerson", so I quite like the suggestion.

test/models/IllPersonSchema.test.ts
akaJSaunders: Im sure we can do both, I` prefer should reject empty object with only one error`, maybe rename the 'rejects' description to something else?

src/models/Illness.ts
akaJSaunders: Ah i get it, okay thats fine :)

src/models/fields/IllPerson.schema.ts
akaJSaunders: make sure to double check it all works fine

src/views/illness/ill-person.njk
```diff
@@ -1,16 +1,95 @@
...
+        }) if validationResult and validationResult.errors.length > 0
+      }}
+
+      <form method="post">
+        {% set otherPersonHtml %}
+          {{ govukInput({
+            id: "other-person",
+            name: "otherPerson",
+            type: "text",
+            spellcheck: false,
```
akaJSaunders, line 30: is this possibly false by default? I don't think we have this spellcheck field in any other template

test/models/IllPersonSchema.test.ts
rhysbarrett: How's this? :) Think "invalid" and "valid" look more like subtitles than bad sentence structure so this works.

Screenshot 2020-08-04 at 10 46 20

src/views/illness/ill-person.njk
rhysbarrett: The design system page doesn't specify the default value, but you're right, this doesn't seem necessary either way. Yet another artefact of poor copypasting. I'll be more diligent in future.

test/models/IllPersonSchema.test.ts
akaJSaunders: Im going to hang that up my wall <3

src/controllers/IllPersonController.ts
```diff
@@ -1,17 +1,89 @@
...
+
+    protected prepareSessionModelPriorSave(appeal: Appeal, value: FormBody): Appeal {
+
+        /* const dummyOther: OtherReason = {
+            title: 'Dummy Reason',
+            description: 'The current Appeal data model requires an Other-type reason, ' +
+                'and reworking the data model is outside of the scope of this feature. ' +
+                'Until the Appeal object has been remodelled, this dummy reason must ' +
+                'be included.'
+        }; */
```
akaJSaunders, line 71: Please remove

src/controllers/IllPersonController.ts
```diff
@@ -1,17 +1,89 @@
...
+
+    protected prepareViewModelFromAppeal(appeal: Appeal): any {
+        const illness: Illness | undefined = appeal.reasons?.illness;
+        if (!illness) {
+            return {};
+        }
+
+        return {
+            illPerson: illness.illPerson,
+            otherPerson: illness.otherPerson
```
akaJSaunders, line 59: someoneElse to keep it consistent

src/models/Illness.ts
```diff
@@ -1,7 +1,10 @@
...
 
 export interface Illness {
+    illPerson: IllPerson;
+    otherPerson?: string;
```
akaJSaunders, line 7: someone else to keep it consistent?

src/models/fields/IllPerson.ts
```diff
@@ -0,0 +1,7 @@
+export enum IllPerson {
+    director = 'director',
+    accountant = 'accountant',
+    family = 'family',
+    employee = 'employee',
+    someoneElse = 'someoneElse'
```
akaJSaunders, line 6: You are using otherPerson in a number of other places in your code, I think its worth going over those locations and trying to refactor.

src/views/illness/ill-person.njk
```diff
@@ -1,16 +1,95 @@
...
+          {{ 
+            govukInput({
+              id: 'other-person',
+              name: 'otherPerson',
+              type: 'text',
+              classes: 'govuk-!-width-one-third',
+              label: {
+                text: 'Their relationship to the company'
+              },
+              errorMessage: validationResult.getErrorForField('otherPerson') if validationResult
```
akaJSaunders, line 34: refactor to someone else pehaps

src/views/illness/ill-person.njk
```diff
@@ -1,16 +1,95 @@
...
+                value: 'employee',
+                text: 'A company employee',
+                checked: true if illPerson === 'employee'
+              },
+              {
+                value: 'someoneElse',
+                text: 'Someone else',
+                checked: true if illPerson === 'someoneElse',
+                conditional: {
+                  html: otherPersonHtml
```
akaJSaunders, line 77: rename html to someoneElseHtml

src/controllers/IllPersonController.ts
rhysbarrett: Ah yes, duh, thank you.

PR #265, opened 2020-08-05
feat: Choose Reason transitions

rhysbarrett:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1901

### Change description

This commit implements and tests navigation forwards and backwards from the "Choose an Appeal Reason" page:

- Clicking "Continue" leads to the first screen of the appropriate journey, depending on which reason's radio button was selected.
- Clicking "Back" leads to the "Review Penalty" page.

Continue functionality is achieved through a new middleware `ReasonRouter`, and a new (_optional_ so as not to cause issues elsewhere in the codebase) session object variable `currentReasonType`. The new variable is intended as a _temporary_ method of tracking the user's current reason type, until the Appeal data structure is upgraded later.

### Work checklist

### Merge instructions

```
```


src/controllers/ChooseAppealReasonController.ts
```diff
@@ -26,10 +30,23 @@ const navigation: Navigation = {
...
 };
 
-@controller(CHOOSE_REASON_PAGE_URI, FeatureToggleMiddleware(Feature.ILLNESS_REASON), SessionMiddleware, AuthMiddleware)
-export class ChooseAppealReasonController extends BaseController<any>{
+interface FormBody {
+    reason: ReasonType;
+}
+
+@controller(CHOOSE_REASON_PAGE_URI, FeatureToggleMiddleware(Feature.ILLNESS_REASON), SessionMiddleware, AuthMiddleware,
+CompanyAuthMiddleware)
```
n-r-b, line 38: can you please update the `AppliedCompanyAuthMiddleware.test.ts` with this page

src/middleware/ReasonRouter.ts
```diff
@@ -0,0 +1,51 @@
...
+        const applicationData: ApplicationData | undefined = session.getExtraData(APPLICATION_DATA_KEY);
+        if (!applicationData) {
+            throw APPLICATION_DATA_UNDEFINED;
+        }
+
+        const reason = applicationData.appeal.currentReasonType;
+        let redirect: string | undefined;
+
+        switch(reason) {
+            case ReasonType.illness:
```
n-r-b, line 37: can we not just add the routing logic to where we're building the navigation control in the `ChooseAppealReasonController.ts`? It no longer makes sense to navigate to the OtherReasonDisclaimer page by default

src/controllers/ChooseAppealReasonController.ts
rhysbarrett: Ah yes, good spot.

src/middleware/ReasonRouter.ts
rhysbarrett: I agree - sending the user to the Other Reason journey by default doesn't sit well with me - but I could work out how to do this within the navigation definition. I couldn't work out how to gain access to the session (or better yet, the form data, so that we don't need this new session variable) from there. Is this something we can do??

rhysbarrett: Resolved as suggested.

rhysbarrett:
3 Amigos suggested but skipped.

PR #266, opened 2020-08-05
fix: calendar invalid dates should trigger validation error

akaJSaunders:
### Change description

Example of dates that should be invalid:
31st of Feb
30th of Feb
31st of April

### Work checklist

### Merge instructions

```
```


rhysbarrett:
Looks and works great! Only thing I'd suggest is a test for 29th Feb in a leap year and a non-leap year, because if the validation can handle that (it can, I manually tested it just now, but an automated test would be even better), then it should be able to handle anything.

PR #267, opened 2020-08-05
add: transition functionality from Who is Ill page

akaJSaunders:
### JIRA link

https://companieshouse.atlassian.net/browse/LFA-1443

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/IllPersonController.ts
+++ b/src/controllers/IllPersonController.ts
@@ -12,17 +12,17 @@ import { OtherReason } from 'app/models/OtherReason.ts';
 import { IllPerson } from 'app/models/fields/IllPerson';
 import { schema } from 'app/models/fields/IllPerson.schema';
 import { Feature } from 'app/utils/Feature';
-import { ILL_PERSON_PAGE_URI } from 'app/utils/Paths';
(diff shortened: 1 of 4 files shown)


src/controllers/IllPersonController.ts
```diff
@@ -61,11 +62,21 @@ export class IllPersonController extends BaseController<FormBody> {
...
         if (appeal.reasons?.illness != null) {
+
             appeal.reasons.illness.illPerson = value.illPerson;
+
             if (value.illPerson !== IllPerson.someoneElse) {
+
                 appeal.reasons.illness.otherPerson = undefined;
+
+            } else {
+
```
rhysbarrett, line 75: I'm curious about all of the additional whitespace, particularly in this `else` block. Is there a style guide I need to take a look at?

akaJSaunders: I only did that because when debugging I was finding it hard to read. I think its subjective.

rhysbarrett: Fair enough! Just checking :)

akaJSaunders: Im not very consistent, but since we had nested if and else, I thought it might be better

PR #289, opened 2021-02-11
Fix/revert lockfile to v1

mikecsmith:
Reverts lockfile to v1 - project currently does not build in our `CI` platform due to them running `npm@6` however recent updates to this project use a lockfile created with `npm@7` which causes issues with the `make package` command.

This commit has deleted the `package-lock.json` and recreated it using `npm@6`

It has also stopped typechecking of libraries to fix an issue with `kafka-node` - please see https://github.com/SOHU-Co/kafka-node/issues/1412 for more details.


mouhajer-ch (approved):
Thanks Mike! Tested the build and it works absolutely fine :+1:

PR #302, opened 2021-07-06
Feature/bi 8248 remove embedded reasons logic on view pages

mouhajer-ch:
### JIRA link
Part of [BI-8248](https://companieshouse.atlassian.net/browse/BI-8248)

### Change description
Remove embedded reasons logic on view pages

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/OtherReasonController.ts
+++ b/src/controllers/OtherReasonController.ts
@@ -32,7 +32,7 @@ export class OtherReasonController extends SafeNavigationBaseController<OtherRea
     }
 
     protected prepareViewModelFromAppeal(appeal: Appeal): Record<string, any> & OtherReason {
-        return appeal.reasons?.other;
+        return appeal.reasons?.other as OtherReason;
     }
 
     protected prepareSessionModelPriorSave(appeal: Appeal, value: OtherReason): Appeal {
--- a/src/views/check-your-appeal.njk
+++ b/src/views/check-your-appeal.njk
@@ -30,15 +30,15 @@
           rows: [
             {
               key: { text: 'Company Name' },
-              value: { text: penaltyIdentifier.companyName }
+              value: { text: appealPenaltyDetails.companyName }
             },
             {
               key: { text: 'Company Number' },
-              value: { text: penaltyIdentifier.companyNumber }
+              value: { text: appealPenaltyDetails.companyNumber }
             },
             {
               key: { text: 'Type' },
-              value: { text: 'Late Filing Penalty (' + penaltyIdentifier.penaltyReference + ')' }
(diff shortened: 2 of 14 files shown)


src/controllers/OtherReasonController.ts
```diff
@@ -32,7 +32,7 @@ export class OtherReasonController extends SafeNavigationBaseController<OtherRea
     }
 
     protected prepareViewModelFromAppeal(appeal: Appeal): Record<string, any> & OtherReason {
```
CH-JonathanBlake, line 34: could update return types.

src/views/confirmation.njk
```diff
@@ -47,9 +47,9 @@
         <h2 class="govuk-heading-l">Reason for appeal</h2>
 
         {% set attachmentsList %}
```
CH-JonathanBlake, line 49: After setting nothing(?) to this variable is is then not used.
I wonder if the intention was 
`{% set attachmentsList = appealReasonDetails.attachments %}`
https://mozilla.github.io/nunjucks/templating.html#set

src/views/check-your-appeal.njk
```diff
@@ -51,11 +51,11 @@
       <h2 class="govuk-heading-l">Reason for appeal</h2>
 
       {% set attachmentsList %}
```
CHdlewis, line 53: Check this?  Potentially looking at an array

src/controllers/OtherReasonController.ts
CH-JonathanBlake: When does it return `Record<string, any>`?

mouhajer-ch: The [recordkeystype](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype) for how has been implemented through all the repo is just telling us that the object has index signature of type string. The important part is the second object type, the first I would say is just a reinforce type

src/views/confirmation.njk
mouhajer-ch: **attachmentsList** is used as variable for the supporting document where the html content created inside it is just an ordered list if `appealReasonDetails.attachments` exists otherwise a simple `None string.
Used as an alternative for macros to have a clean njk page

src/views/check-your-appeal.njk
mouhajer-ch: **attachmentsList** is used as variable for the supporting document where the html content created inside it is just an unordered list if `appealReasonDetails.attachments` exists otherwise a simple `None string.
Used as an alternative for macros to have a clean njk page

CHdlewis (approved):
Looks good

PR #356, opened 2021-09-17
Add validation to illness end date

mreis-ch:
[BI-8946](https://companieshouse.atlassian.net/browse/BI-8946)

Create new validator class IllnessEndDateValidator. This extends DateValidator and includes functionality to check whether the end date is before the start date.
Update IllnessEndDateController to use new validator class, IllnessEndDateValidator.
Update start date in future error message to date in future error message to make now generic

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/IllnessEndDateController.ts
+++ b/src/controllers/IllnessEndDateController.ts
@@ -1,9 +1,9 @@
 import { SessionMiddleware } from 'ch-node-session-handler';
 import { controller } from 'inversify-express-utils';
 import moment from 'moment';
+import { IllnessEndDateValidator } from './validators/IllnessEndDateValidator';
 
 import { SafeNavigationBaseController } from 'app/controllers/SafeNavigationBaseController';
-import { DateValidator } from 'app/controllers/validators/DateValidator';
 import { AuthMiddleware } from 'app/middleware/AuthMiddleware';
 import { FeatureToggleMiddleware } from 'app/middleware/FeatureToggleMiddleware';
 import { loggerInstance, loggingMessage } from 'app/middleware/Logger';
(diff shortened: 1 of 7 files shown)


src/controllers/IllnessEndDateController.ts
```diff
@@ -1,9 +1,9 @@
 import { SessionMiddleware } from 'ch-node-session-handler';
 import { controller } from 'inversify-express-utils';
 import moment from 'moment';
+import { IllnessEndDateValidator } from './validators/IllnessEndDateValidator';
```
mouhajer-ch, line 4: For consistency better use the src application path 
`import { IllnessEndDateValidator } from 'app/controllers/validators/IllnessEndDateValidator';`

src/controllers/validators/IllnessEndDateValidator.ts
```diff
@@ -0,0 +1,60 @@
...
+        const validationResult: ValidationResult = await super.validate(request);
+
+        const session: Session | undefined = request.session;
+
+        if (!session) {
+            throw SESSION_NOT_FOUND_ERROR;
+        }
+
+        const appData: ApplicationData = session.getExtraData<ApplicationData>(APPLICATION_DATA_KEY)
+            || { appeal: {} as Appeal, navigation: { permissions: [] } } as ApplicationData;
```
mouhajer-ch, line 41: if appData is undefined I will throw an Error instead, something like

    if(!appData) {
        throw APPLICATION_DATA_UNDEFINED;
    }

src/controllers/IllnessEndDateController.ts
mreis-ch: Ah yes, ok 👍

src/controllers/validators/IllnessEndDateValidator.ts
mreis-ch: Will do

mreis-ch: Just want to check. Line 40, doesn't this line always set appData? I'm guessing if the getExtraData returns nothing then an empty appData object is created?

mouhajer-ch: At this page `appData` should not be empty, better fail quick

mouhajer-ch (approved):

PR #480, opened 2025-07-14
feat: add `ACCOUNT_PRIVATE_URL` to companyAuthConfig

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/CC-2621

### Change description

Adds `ACCOUNT_PRIVATE_URL` to companyAuthConfig

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -10,6 +10,7 @@ DEFAULT_SESSION_EXPIRATION=3600
 
 ACCOUNT_WEB_URL=http://web.chs-dev.internal:4000
 ACCOUNT_URL=http://account.chs.local
+ACCOUNT_PRIVATE_URL=http://private.account.chs.local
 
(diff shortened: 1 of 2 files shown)


.env.test
```diff
@@ -10,6 +10,7 @@ DEFAULT_SESSION_EXPIRATION=3600
 
 ACCOUNT_WEB_URL=http://web.chs-dev.internal:4000
 ACCOUNT_URL=http://account.chs.local
+ACCOUNT_PRIVATE_URL=http://private.account.chs.local
```
NeilShadrach, line 13: Where does this url come from?
If it is just a dummy value maybe private-account.chs/local would be more consistent with the URLs used in other environments

NeilShadrach (approved):

